Issue #334, opened 2013-06-20
async parallel error handling


MichaelMcKavanagh:
When processing multiple functions in parallel, using async parallel, if one of those functions throws an error via it's callback, async parallel processes it's callback before some of it's functions have completed.  This may be the desired behaviour under some situations, however, I desire different logic.  I would like all functions to run regardless of if some of them error or not, and then to be able to process through an error array after they have all completed processing.

My workaround for this situation currently is for my functions running in parallel to always not throw an error via normal way, but instead pretend they succeeded, and pass their error/data through an object in the result.  In this way all of the functions will run, and then I have to manually process through the result objects looking for and handling any errors.  

However, bypassing normal error handling in order to workaround async parallel isn't ideal in my mind.  Perhaps there could be a better solution to provide either option for us, depending on the desired behaviour.

PR #349, opened 2013-07-15
parallelCollect method.

daviddenton:
A solution to issue #334 - regarding async parallel error handling. Added new method, tests and documentation for it.

Generally, this is a version of parallel() which, instead of failing on the first error, continues and gathers all errors. The final callback is called with the error list/object and the result list/object.

I've now implemented this functionality across a few projects - it seems to be something that would be useful in the core.


aearly:
Reopening this for reference, because I think it's the best  implementation of this feature.

aearly:
Closing this in favor of #942.  There are several things to figure out in order to support this type of behavior, then we can re-implement.

Issue #299, opened 2013-04-19
IE 10 async.setImmediate and async.nextTick cause "Invalid calling object"


jsas:
setImmediate and nextTick work fine on their own, but IE 10 doesn't like setImmediate being run as a property of async. async.setImmediate.call(null, arg) does work, though. Or I guess you could make nextTick and setImmediate vars in the scope of async and call them directly.

To reproduce: In an IE 10 console, run this: 

({setImmediate:setImmediate}).setImmediate(function(){})

A quick and dirty way to set them would be:

async.setImmediate = setImmediate.bind(null);
async.nextTick = setImmediate.bind(null);

I opted for this:

``` javascript
//// nextTick implementation with browser-compatible fallback ////
    if (typeof process === 'undefined' || !(process.nextTick)) {
        if (typeof setImmediate === 'function') {
            try {
                async.setImmediate = async.nextTick = setImmediate;
                try {
                    async.setImmediate(function(){});
                } catch(e) {
                    async.setImmediate = async.nextTick = setImmediate.bind(null);
                }
            } catch(e) {
                async.setImmediate = null;
            }
        }
        if (!async['setImmediate']) {
            async.nextTick = function (fn) {
                setTimeout(fn, 0);
            };
            async.setImmediate = async.nextTick;
        }
    }
    else {
        async.nextTick = process.nextTick;
        async.setImmediate = setImmediate;
    }
```

PR #350, opened 2013-07-16
Fix setImmediate incompatibility with IE10 even when process and process.nextTick are defined

louischatriot:
Hello,

This references issue https://github.com/caolan/async/issues/299 . The fix sent by @dougwilson (PR https://github.com/caolan/async/pull/317) only works if process or process.nextTick are not defined. This effectively assumes that in a browser context, process.nextTick is not defined, but that's not always the case. For example, if you use browserify to create a browser version of your code, process.nextTick can be defined and in that case IE10 raises the "Invalid calling object" error. That is the case with one of my projects (the javascript database NeDB, https://github.com/louischatriot/nedb).

So I just copy-pasted @dougwilson's fix in the right section and the issue is solved.

Cheers,
Louis

### DIFF
--- a/lib/async.js
+++ b/lib/async.js
@@ -91,7 +91,10 @@
     else {
         async.nextTick = process.nextTick;
         if (typeof setImmediate !== 'undefined') {
-            async.setImmediate = setImmediate;
+            async.setImmediate = function (fn) {
+              // not a direct alias for IE10 compatibility
+              setImmediate(fn);
+            };
         }
         else {
             async.setImmediate = async.nextTick;


diversario:
No, nothing overrides `setImmediate` anywhere in my code. Perhaps some modules I uses (sinon?..).

dougwilson:
sinon overrides `setImmediate`, `setTimeout`, etc. if you use the fake timers feature. You can find various issues in the Sinon tracker: https://github.com/cjohansen/Sinon.JS/issues?q=setimmediate

dougwilson:
@diversario , also, see https://github.com/cjohansen/Sinon.JS/issues/561

diversario:
Thanks, @dougwilson, that helps!

PR #366, opened 2013-08-22
Added cargoQueue function.

garrows:
The same as cargo but allows concurrency like queue does.


aearly:
We can get this functionality for free now, with the unified `queue` and `cargo` logic.  Just needs docs and tests.

celalo:
@aearly how can this be done? Could you give some hints?

PR #371, opened 2013-09-04
Add a clear function for async.queue()

sundq:
Add a clear function for async.queue(),so user can clear the task queue when they need.

### DIFF
--- a/README.md
+++ b/README.md
@@ -939,8 +939,8 @@ methods:
   alter the concurrency on-the-fly.
 * push(task, [callback]) - add a new task to the queue, the callback is called
   once the worker has finished processing the task.
-* clear() - clear all the task in task queue.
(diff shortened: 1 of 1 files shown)


gregorleban:
Can I ask why is this issue marked as closed although the queue doesn't have the clear() method? Is there a proper way to clear the async.queue?

hargasinski:
@gregorleban have you looked into the [`kill`](http://caolan.github.io/async/docs.html#.QueueObject) function for `async.queue`? It clears the queue, and removes the `drain` callback.

gregorleban:
I  did. that is what I'm currently using but it seems unoptimal since after killing the queue I need then to create a new identical queue and set up the task processing callbacks which seems redundant. It's not a big deal, I was just wondering why the clear() method appeared for a while in the API and was then removed.

PR #396, opened 2013-10-22
Added exception handling to async waterfall method.

diogogmt:
Adding exception handling for async.waterfall

### DIFF
--- a/lib/async.js
+++ b/lib/async.js
@@ -492,7 +492,11 @@
                         args.push(callback);
                     }
                     async.setImmediate(function () {
-                        iterator.apply(null, args);
+                        try {
+                            iterator.apply(null, args);
+                        } catch (e) {
+                            callback(e);
+                        }
                     });
                 }
             };
--- a/test/test-async.js
+++ b/test/test-async.js
@@ -90,6 +90,7 @@ exports['applyEach'] = function (test) {
     var one = function (val, cb) {
         test.equal(val, 5);
         setTimeout(function () {
+
             call_order.push('one');
             cb(null, 1);
         }, 100);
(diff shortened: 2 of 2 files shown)


coderbuzz:
Sorry to bumb it, I'm looking this feature too. +1
I'm too lazy writing try {} catch on each function :p

wookieb:
@caolan @aearly What need to happen to convince you to introduce try/catch error handling for async?

1) Handling nested async errors
I think it's clear it's impossible to do. The same as it's impossible to do for promises.
But the simplest possible case is still a big improvement. Wrapping all possible workers with try/catch (or having helper for that) introduce a lot of boilerplate. As programmers we have to be productive and boilerplate is very counter productive. In that case is also very error-prone.

2) Such huge change requires major version change
Let's do it then instead forcing 1 000 000 + of programmers to introduce a lot of boilerplate because we don't want to do it.

3) Performance hit
Do we have exact measures on that? I'm happy to see them if possible.
If we don't have them I can do it. Just let me know.

4) Async functions should pass errors with callbacks
That's right. Promises should be always resolved or rejected, yet auto rejection on throw has been added to the spec of Promises/A+. Design ideology <<< Reality.

aearly:
Automatically wrapping everything in try/catch internally is actually impossible to do properly.  You either end up catching errors later in the callback chain after a successful task, or you end up having to introduce a deferral for each async step (as promises do).  We just eliminated all unnecessary deferrals throughout Async, so we're not going to add them back in.

wookieb:
Thanks for your quick answer @aearly 

"You either end up catching errors later in the callback chain after a successful task"
Could you please provide some example?

aearly:
```js
function asyncMethod(args, iteratee, callback) {
  try {
    iteratee(args, callback);
  } catch (err) {
    callback(err);
  }
}

asyncMethod(args, asyncFn, function (err, res) {
  throw new Error("in callback"); <-- this error will be caught above
})

```

Issue #508, opened 2014-04-14
async.queue pause/resume problem


i3fox:
Async v.0.7.0

In the async.queue I found an issue related to restart after a pause (resume) and I try to fix.
It's not simple to explain, please follow these steps:
1) for concurrency tasks are in execution
2) I pause the queue and wait for the execution of tasks finishes
3) I resume queue
4) Queue restarts but concurrency does not work: only one task is running.

Below you find 
a) Example of problem
b) My purpose to fix

please, let me known if it's ok.
Best regards 
Fox

Example of problem:

```
var async = require("async");
var q = async.queue(function (task, callback) {
    console.log('hello ' + task.name);
    setTimeout(function() {
      //console.log('bye ' + task.name);
      callback();
    }, 2000);
}, 5);

q.pause();
q.push({name: '1'});
q.push({name: '2'});
q.push({name: '3'});
q.push({name: '4'});
q.push({name: '5'});
q.push({name: '6'});
q.push({name: '7'});
q.push({name: '8'});
q.push({name: '9'});
q.push({name: '10'});

setTimeout(function () {
  q.resume();
}, 2500);
```

My purpose to fix at row number 806 and next

```
            pause: function () {
                if (q.paused === true) { return; }
                q.paused = true;
                // modify
                // before
                //q.process();
            },
            resume: function () {
                if (q.paused === false) { return; }
                q.paused = false;
                // modify
                // before
                //q.process();
                // after
                _each(q.tasks, function(task) {
                    async.setImmediate(q.process);
                });
            }
```

aearly:
Moving discussion to #512.

ORESoftware:
was this resolved?

PR #512, opened 2014-04-16
async.queue pause/resume concurrency problem

i3fox:
In the async.queue I found an issue related to restart after a pause (resume) and I try to fix.
It's not simple to explain, please follow these steps:
1) for concurrency tasks are in execution
2) I pause the queue and wait for the execution of tasks finishes
3) I resume queue
4) Queue restarts but concurrency does not work: only one task is running.

Example of problem:

```javascript
var async = require("async");
var q = async.queue(function (task, callback) {
    console.log('hello ' + task.name);
    setTimeout(function() {
        callback();
    }, 2000);
}, 5);

q.pause();
q.push({name: '1'});
q.push({name: '2'});
q.push({name: '3'});
q.push({name: '4'});
q.push({name: '5'});
q.push({name: '6'});
q.push({name: '7'});
q.push({name: '8'});
q.push({name: '9'});
q.push({name: '10'});

setTimeout(function () {
    q.resume();
    }, 2500);
```

### DIFF
--- a/lib/async.js
+++ b/lib/async.js
@@ -806,12 +806,13 @@
             pause: function () {
                 if (q.paused === true) { return; }
                 q.paused = true;
-                q.process();
             },
             resume: function () {
                 if (q.paused === false) { return; }
                 q.paused = false;
-                q.process();
+                _each(q.tasks, function(task) {
+                    async.setImmediate(q.process);
+                });
             }
         };
         return q;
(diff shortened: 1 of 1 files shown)


elliotbonneville:
Looks like this bug is (still?) appearing in NPM's package, v0.9.0. Any updates since then?

ralgara:
Still an issue in 0.9.0. In #508, @caolan suggests against calling the callback synchronously. I suppose this means wrapping the callback in either setImmediate or process.nextTick. I have tried both to no avail. Any intention to merge this pull request?

aearly:
I just added a couple tests for this and it seems to work fine.  Am I testing what is causing the issue properly?

PR #516, opened 2014-04-22
Add capability to try to detect a possible stack overflow and queue functions the next tick

CrispyConductor:
There are a number of cases where various iterator functions are used in cases where the iterator isn't actually asynchronous (this is usually when the iterator is conditionally synchronous, but may return immediately in some cases).  When iterating over large sets of objects, this can cause unexpected stack overflows.

This commit tries to track estimated stack depth by incrementing a counter each time a tracked function is executed, and clearing the counter each tick.  If the counter grows above a threshold, additional functions are queued to execute the next tick automatically.

This also adds the functions async.stackSafe(callback), async.wrapStackSafe(func), and safe variants of existing functions like safeEach(), safeSeries(), etc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ a method of another library isn't working as an iterator, study this example:
 // Here is a simple object with an (unnecessarily roundabout) squaring method
 var AsyncSquaringLibrary = {
   squareExponent: 2,
-  square: function(number, callback){ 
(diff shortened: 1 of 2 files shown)


aearly:
FYI `ensureAsync` in v1.2.0 and later is a nice way to make an async function "stack-safe".

PR #517, opened 2014-04-24
Return array of errors in parallel forEach (keeping compatibility)

fmenezes:
Keeping compatibility with the callback(err[, errs]), return first error and array of errors.


aearly:
I really dislike the global `collecting` state  here -- it will cause issues with multiple collecting functions running in parallel.  Closing in favor of #349.

PR #527, opened 2014-05-07
Better support for require.js

trusktr:
I'm using Require.js, so `this` doesn't refer to the `window` object, so async didn't work until I made this change.

### DIFF
--- a/lib/async.js
+++ b/lib/async.js
@@ -14,7 +14,16 @@
     // global on the server, window in the browser
     var root, previous_async;
 
-    root = this;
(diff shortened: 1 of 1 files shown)


trusktr:
Updated, finally.

beaugunderson:
@trusktr Note that referencing an undeclared variable directly gives a `ReferenceError`, that's why the tests are failing. You'll want `typeof window !== 'undefined'` to check declaration without a `ReferenceError`. :)

trusktr:
There we go.

trusktr:
bump :)

PR #531, opened 2014-05-13
Add anyway version of map and parallel

allevo:
Use mapAnyway or parallelAnyway for doing something in parallel but if some tasks fail, the others will be executed anyway. In the final callback, the first parameter is valued with all errors. For mapAnyway, the second parameter with all mapped object.
Add main test for parallelAnyway and mapAnyway.

### DIFF
--- a/lib/async.js
+++ b/lib/async.js
@@ -240,7 +240,7 @@
             callback(err, results);
         });
     };
-    var _asyncMapAnyway = function (eachfn, arr, iterator, callback) {
+    var _asyncMapContinue = function (eachfn, arr, iterator, callback) {
         var results = [];
         var errs = [];
         arr = _map(arr, function (x, i) {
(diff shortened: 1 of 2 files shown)


aearly:
I like this idea, but I dislike the global `collect` and `uncollect` states.  Closing this in #349.  We'll think up a general solution.

aearly:
Nevermind, got PRs  mixed up.

kmgilbert100:
+1

gocoffeecup:
+1

kmgilbert100:
Why hasn't this been added yet?

allevo:
Because @caolan hasn't decided the names yet. As soon as possible, I'll resolve the conflict and repush the PR

aearly:
I realized that this breaks one of `map`'s guarantees -- that the results returned in the same order as their inputs in the original array.  If half your tasks callback an error, you'll get a results array that is half the  size of the original array.  It also is a bit of a convention break to return an array of errors (rather than a single `Error` object) as the first arg to a callback.

We're going to go back to the drawing board with #942

PR #535, opened 2014-05-20
Fixed nasty callback duplication.

martinheidegger:
I had some case where the second error response was called during the callback of the first error response in an async call before the callback could be replaced with a noop. That resulted in the error case where the callback was called twice with an error message. That was pretty hard to debug. I figured that the same error could occur at other places as well and changed it there too. It would be nice if someone could provide unit tests for these changes, else I will provide them as soon as I have time for it.

### DIFF
--- a/lib/async.js
+++ b/lib/async.js
@@ -126,8 +126,9 @@
         });
         function done(err) {
           if (err) {
-              callback(err);
+              var cb = callback;
               callback = function () {};
+              cb(err);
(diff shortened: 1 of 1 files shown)


aearly:
I fixed a lot of these issues recently. In most cases all it took was wrapping the callback with `_once`.  If these issues resurface, please open a new PR with the specifics and a test case.

PR #546, opened 2014-06-09
Noop

Neamar:
Using one single `noop` var, we can avoid creating empty function on the fly and gain (some) perf.

This is not much, but since it cost nothing and adheres to common node.js standard, i thought this was worth a little PR.


Neamar:
(@beaugunderson rebased to work on master)

PR #551, opened 2014-06-12
Added stop function to queue

bouddhalee:
I created a stop function to help me with that scenario:
   - Stop an action that used a queue for running subtasks.
   - Know when all the running subtasks have completed.

The current kill method removes the drain callback as it should, so I added a stop function that will keep the drain.
Stop will also make sure that the drain is called if there is no active worker. I did that in case stop is called before the workers started running.

### DIFF
--- a/lib/async.js
+++ b/lib/async.js
@@ -785,6 +785,14 @@
               q.drain = null;
               q.tasks = [];
             },
+            stop: function () {
+                // call drain right away if no worker is running
+                // else it will be called by the last worker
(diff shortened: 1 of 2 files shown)


aearly:
I'd be :+1: for this if it had docs.  It fits nicely between `pause` and `kill`.

aearly:
`queue` has changed a lot since this PR.  You can submit a new PR with this feature if you still need this.  (Along with docs and tests)

PR #553, opened 2014-06-14
cargoDelayed

brugnara:
I wrote this because I needed to scheduling a cargo departure. This allow me to receive from my webserver lots of commands and execute when needed, which is not now :smile: 
Hope you like.

Ps: my editor removed any trailing space on line end. If this is a problem, please let me know and I'll redo.
Best regards.
Daniele.


aearly:
I feel like this is best accomplished by:

```js
var delayedCargo = async.cargo(worker, payload);

delayedCargo.delayedPush = function (task, callback) {
  setTimeout(function () {
    delayedCargo.push(task, callback);
  }, 3*60*1000);
}
```

PR #558, opened 2014-06-17
Defense for eached array alteration, fixes #557

### LINKED ISSUE #557: async.each callback could be called more than once when array is modified
```javascript
var async = require('async')

inputs = [1,2]
async.each(inputs,
	function (item, callback) {
		setTimeout(function(){callback()}, 1000)
	},
	function(err){
		console.log("async.each done")	
	})

inputs.splice(1,1)

/* output is:
async.each done
async.each done
*/
```

the problem is async use completed >= arr.length to check if call callback. however arr.length is less than async.each was called.

amv:
Altering the array of iterables after calling each on the array makes async either call the callback several times or not call the callback at all (depending on wether you add or remove array items).

Although it is probably not a very good practise to alter the array while it is being iterated with each, sometimes the array is passed on and the person receiving the array does not know that such requirement is in place.

The current async behaviour in such situations is very problematic and this commit should fix the problems. This should fix #557 .

### DIFF
--- a/lib/async.js
+++ b/lib/async.js
@@ -121,6 +121,7 @@
             return callback();
         }
         var completed = 0;
+        var arr_length = arr.length;
         _each(arr, function (x) {
             iterator(x, only_once(done) );
         });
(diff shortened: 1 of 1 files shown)


aearly:
This was fixed as side effects of other changes.

PR #560, opened 2014-06-18
Added a timesLimit() function

chad3814:
like times() and timesSeries() it is just a passthrough for mapLimit()


aearly:
Added in f1263b39

PR #561, opened 2014-06-19
improve cross-browser functions

jmgunn87:
This PR removes an 'if' statement from each cross browser functions. This is a micro optimisation but it can't hurt.


aearly:
Fixed through 3ffbf2b.

PR #562, opened 2014-06-20
go 1.0 + comply with semver

Raynos:
semver says

> Major version zero (0.y.z) is for initial development. Anything may change at any time. The public API should not be considered stable.

This means that 0.9.1 could break back compat of async, which is very sad.

`async` seems to be a rock solid production quality piece of software.

It's ready to go 1.0 and comply with semver, i.e. all new additions to the interface bump minor, and any breaking changes to the interface will bump major.


aearly:
Version 1.0.0 published!

PR #570, opened 2014-07-03
Arg validation in .waterfall() is now asynchronous

jsdevel:
I noticed that the argument validation in `.waterfall()` would call the cb right away.


aearly:
We're going to start avoiding unnecessary deferrals.  See #696

PR #576, opened 2014-07-07
Use all of the arguments in default memoize hasher. Fixes #575

### LINKED ISSUE #575: Default memoize hasher returns invalid key for multiple arguments
Currently, the default memoize hasher takes only the first argument to create the memo key:

https://github.com/caolan/async/blob/master/lib/async.js#L1000-L1002

This will cause memoized functions that take multiple arguments to call back with invalid results:

```javascript
var fn = async.memoize(function (a, b, c, cb) {
  cb(a, b, c);
});

fn(1, 2, 3, function (a, b, c) {
  console.log(a, b, c); 
  // 1, 2, 3 - correct
});

fn(1, 10, 20, function (a, b, c) {
  console.log(a, b, c); 
  // 1, 2, 3 - incorrect -- should be "1, 10, 20"
});
```

lxe:
See #575

### DIFF
--- a/lib/async.js
+++ b/lib/async.js
@@ -998,7 +998,11 @@
         var memo = {};
         var queues = {};
         hasher = hasher || function (x) {
-            return x;
+            if (arguments.length <= 1) {
+                return x;
+            } else {
+                return Array.prototype.slice.call(arguments);
+            }
         };
         var memoized = function () {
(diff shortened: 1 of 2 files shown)


aearly:
Sorry, I opted for a docs change, rather than changing the default behavior of the hasher.  This would have been a breaking change.

Issue #429, opened 2013-12-23
is there any way to set timeout in parallel?


stoneshim:
I want to run multiple functions in parallel, in specified time.
If timeout is occurred, i want to get completed results at the moment. (except the results of uncompleted functions at the moment)
Is there any way to do that?

In my opinion, if i need partial results (completed results of functions before timeout is occurred) when timeout is occurred, setTimeout() must be in async function.

I need function like this...
`async.parallelInTime(tasks, timeout, [callback]);`

What do you think about this?

```
async.parallelInTime({ 
    one: function(callback){ 
        setTimeout(function(){ 
            console.log('1'); 
            callback(null, 1); 
        }, 400); 
    }, 
    two: function(callback){ 
        setTimeout(function(){ 
            console.log('2'); 
            callback(null, 2); 
        }, 200); 
    } 
}, 
300, // timeout in mille seconds.
function(err, results) { 
    if(err) { 
        console.log(err); // [Error: Timeout occurred.]
    } 
    console.log(results); // { two: 2 }
}); 
```

PR #587, opened 2014-07-25
Add timeout wrapper for any callback interface

marcooliveira:
This solves https://github.com/caolan/async/issues/429#issuecomment-40054121.

Enables timing out a callback interface by wrapping the function and preventing it from being called after it has timed out.


oshybystyi:
vote for that :+1:

aearly:
You can reopen this if you resolve the  merge conflicts.  Also, the docs don't give a clear example of how this can be used.  Do you wrap the callback, or the whole async function?

PR #600, opened 2014-08-24
Added recursive

arb:
Added `async.recursive` which is similar to `forever` but allows you to pass values into it and via `next`. I'm open to suggestions on the name since it technically never ends like a recursive function.


aearly:
Closing this --  seems easy enough to do:

```js
function recursiveTask(arg, cb) {
  doSomethingAsync(arg, function (err, result) {
    if (err) { return cb(err); }
    recursiveTask(result, cb);
  });
}

recursiveTask("foo", done);
```

PR #604, opened 2014-08-29
Use setImmediate to call eachSeries iterator to prevent stack overflow

jjd314:
Long lists were causing "RangeError: Maximum call stack size exceeded" with the direct call to iterator().


aearly:
Closing based on discussion in #696

PR #608, opened 2014-08-31
autoInject - dependency injection for auto.

ghost:
Dependency injection support for auto, for better readability of task graphs.

```js
async.autoInject({
    template:        function(cb)                         { fs.readFile(templateFilename, cb); },
    serverData:      function(cb)                         { http.get(dataUrl, cb); },
    templateData:    function(cb, serverData)             { formatDataForTemplating(serverData, cb); },
    appliedTemplate: function(cb, template, templateData) { mustache.to_html(template.toString(), templateData, cb); }
}, function(err, appliedTemplate) {
    if (err) return handleError(err);
    Serve(appliedTemplate);
});
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,6 +147,7 @@ Usage:
 * [`priorityQueue`](#priorityQueue)
 * [`cargo`](#cargo)
 * [`auto`](#auto)
+* [`autoInject`](#autoInject)
 * [`retry`](#retry)
 * [`iterator`](#iterator)
 * [`apply`](#apply)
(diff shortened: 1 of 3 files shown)


aearly:
BTW,  I'm liking this more as I think about it.  It'll just have the same minification caveats as Angular.  It'll be really handy in server-side code.

steverobb:
Yeah, server-side and Node scripts was mostly what I was thinking about for this feature.  And `auto` will still exist for code that needs to be minification-aware.

I notice in #957 that you are planning to always have the callback last.  Should I make that change here or were you planning to make it yourself after merging?

aearly:
Yeah, I'd prefer to always have the callback be the last argument.  No sense in doing it one way if we're just going to turn around and change it soon after.

aearly:
Closing in favor of #1055

PR #611, opened 2014-09-02
prevent faking setImmediate call

### LINKED ISSUE #609: Is Async somehow dependent on timers ?
I have run into rather weird issue. I am using SinonJS to write my unit tests. There is some neat utility to [fake timers](http://sinonjs.org/docs/#clock). Today I was totally stumbled when my tests became somewhat "stuck". I run them and nothing happened, process was just hanging there and doing nothing.

I started digging and find out that if I don't use fake timers, it works, but since I need to control time in my tests, I started digging more and found out that Async is probably forcing asynchronous behavior with real timers. Check out this [JSFiddle](http://jsfiddle.net/FredyCr/2tbyvvc4/2/).

It's rather innocent piece of code that is not even asynchronous and you can see resulting message in console. However once you remove comment on first line with fake timers, it doesn't finish.

For now I had to get rid of `waterfall` and it started working as expected. I suppose that nothing can be done about this :(

FredyC:
Fixes #609

### DIFF
--- a/lib/async.js
+++ b/lib/async.js
@@ -105,9 +105,11 @@
     else {
         async.nextTick = process.nextTick;
         if (typeof setImmediate !== 'undefined') {
+            // store reference in closure to prevent faking
+            var setImmediateRef = setImmediate;
             async.setImmediate = function (fn) {
               // not a direct alias for IE10 compatibility
-              setImmediate(fn);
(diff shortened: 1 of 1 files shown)


aearly:
Fixed via e417af6.

PR #620, opened 2014-09-10
Retry without callback

jessehouchins:
Updated the docs and made one minor change (for convenience) to the retry method based on https://github.com/caolan/async/issues/597. The API is the same.

There is one failing test for me (noConflict - node only) when running `npm test` in terminal. It also fails in master.


aearly:
Closing in favor of #941.

Issue #263, opened 2013-03-09
enhance auto(…) to detect missing dependencies 


RubyTuesdayDONO:
i'm not sure whether this is by design or not, but it would aid troubleshooting if `auto(…)` could (optionally?) detect when one of its tasks has unsatisfied dependencies. although developers should certainly be careful, it would be so helpful if `auto(…)` were to call out something like, `task 'unreachable' has unsatisfied dependency on task 'nonexistent'`. 

```
var async = require('async'); 
async.auto({ 
    unreachable: [ 'nonexistent', function(callback) { 
        console.log('this message will not appear'); 
        callback(); 
    } ] 
}, function(err,res) { 
    if(err) { 
        console.dir(err); 
        process.exit(1); 
    } else { 
        process.exit(0); 
    } 
} 
```

PR #663, opened 2014-11-18
Detecting dead-locks in async.auto() and throwing an Error

Mickael-van-der-Beek:
This pull request adds dead-lock detection in ```async.auto()``` due to dependencies and thus fixes caolan/async#263.

The three dead-lock cases that were handled and for which test cases were added are:

- 1.  inexistant dependency; e.g:
```
async.auto({
    task1: ['noexist', function(callback, results){
        callback(null, 'task1');
    }]
});
```
- 2. cyclic dependencies; e.g:
```
async.auto({
    task1: ['task2', function(callback, results){
        callback(null, 'task1');
    }],
    task2: ['task1', function(callback, results){
        callback(null, 'task2');
    }]
});
```
- 3. self as dependency which is a subcase of ```2.```; e.g:
```
async.auto({
    task1: ['task1', function(callback, results){
        callback(null, 'task1');
    }]
});
```

If any of these cases are encountered, a verbose `Error()` is thrown.

It is possible to improve the performance a little bit further by embedding these detections inside the ```ready()``` method because we already loop through the ```requires``` array during the ```Array.reduce()``` call:

[Mickael-van-der-Beek/async@2410077e26d5429ac45533438f008706dcb989f7/lib/async.js#L492](https://github.com/Mickael-van-der-Beek/async/blob/2410077e26d5429ac45533438f008706dcb989f7/lib/async.js#L492)

but I felt that it would be a dirtier solution.
If you prefer the faster solution though, just ask and I'll create a new pull-request for it.


Mickael-van-der-Beek:
Is there anyway for this to be merged ?

If nobody's got time and trusts me enough to give me contributor access, I could do it myself.

aearly:
We have to resurrect @caolan from his slumber.

PR #664, opened 2014-11-20
Fix queue draining, add drain_once()

abortz:
Also fixed a test that was explicitly testing broken behavior
and added a test of drain_once() the demonstrates its usefulness.


aearly:
The queue implementation has changed a lot -- is this still needed?

PR #681, opened 2014-12-06
Dereference internal "results" data structures (memory leak)

chainsawsalad:
Once “results” is passed to the callback, the method no longer needs reference to it. Setting to null will encourage Garbage Collection to clean it up. Because the callback is called from within a closure, GC doesn’t automatically recognize that “results” is no longer needed in this context.

After running numerous Heap Snapshot comparisons over time, it's clear that these methods hang onto these objects well after execution time. If the data being passed into results is very large, like in my case the page sources of URLs, the memory leaks becoming crippling after thousands of iterations.

### DIFF
--- a/lib/async.js
+++ b/lib/async.js
@@ -252,6 +252,7 @@
                 });
             }, function (err) {
                 callback(err, results);
+                results = null;
             });
         }
     };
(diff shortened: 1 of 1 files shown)


aearly:
Closing this, since there are conflicts.  We're putting together memory profiling, I'd like to see an example where memory leaks occur.

PR #168, opened 2012-08-10
Adding forEachOf to Iterate Objects

dominicbarnes:
So, I needed the ability to iterate objects (and get the `key` passed into the iterator) and noticed it was not done because it would break compatibility with the Node.js standard libs. I am proposing this "convention" of adding the `Of` suffix on iterators that deal with objects. (instead of arrays)

If this entire concept is rejected outright, then I'll stop here. However, if this seems like a good idea/convention I can finish up adding the same concept to the other various collection iterators.

Usage:

    async.forEachOf({ a: 1, b: 2 }, function (value, key, callback) {
        // value = 1, 2
        // key   = a, b
    }, function (err) {
        // complete
    });

PR #705, opened 2015-01-20
Completing forEachOf

aearly:
This is mostly @dominicbarnes 's work from #168.  The last missing piece was documentation, which I added.  I also made sure it merges with what is currently on master, and simplified the logic in one place.

### DIFF
--- a/lib/async.js
+++ b/lib/async.js
@@ -280,7 +280,7 @@
 
 
 
-    async.forEachOfLimit = async.eachOfLimit =  function (obj, limit, iterator, callback) {
+    async.forEachOfLimit = async.eachOfLimit = function (obj, limit, iterator, callback) {
         _forEachOfLimit(limit)(obj, iterator, callback);
     };
 
--- a/lib/async.js
+++ b/lib/async.js
@@ -171,64 +171,6 @@
     };
     async.forEachSeries = async.eachSeries;
 
-    async.forEachOf = function (object, iterator, callback) {
-        callback = callback || function () {};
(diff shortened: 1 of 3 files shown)


dominicbarnes:
Nice! Would be great to finally see this make it in there!

lib/async.js
```diff
@@ -68,6 +68,14 @@
         return memo;
     };
 
+    var _forEachOf = function (object, iterator) {
+        for (key in object) {
```
Adezandee: May i ask why you don't use Object.keys() ?

test/test-async.js
```diff
@@ -2806,20 +2999,20 @@ exports['cargo bulk task'] = function (test) {
 };
 
 exports['cargo drain once'] = function (test) {
-   
+
```
beaugunderson: I'm personally a fan of more vertical whitespace but these don't match the rest of the file and they seem extraneous to the work done in this PR. :)

aearly: These are trailing whitespace, no lines deleted.  My editor automatically does that.

beaugunderson:
Oh oops, I misread. Totally fine then. :)

lib/async.js
aearly: No reason, I kept the original implementation.  I was focused more on merge conflicts.   `_keys` would be faster here...

mjhasbach:
+1

WillFrew:
+1

felipou:
+1 (when will github add issue voting?)

PR #707, opened 2015-01-28
Add try-catch on every user iterator, task, etc.

IvanDimanov:
Hi there,

You must be hearing this a lot but I really enjoy working with this great lib for a long time and I'm happy there are great guys like you to built it.

In this singe commit I put a try-catch on every user function that can throw and send the caught error to the callback function.
This will make the async lib much safer and predictable for users.

All tests are passing, hope you enjoy it.

### DIFF
--- a/lib/async.js
+++ b/lib/async.js
@@ -119,7 +119,11 @@
         }
         var completed = 0;
         _each(arr, function (x) {
-            iterator(x, only_once(done) );
+            try {
+                iterator(x, only_once(done) );
+            } catch (error) {
+                done(error);
+            }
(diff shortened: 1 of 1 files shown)


aearly:
This is kind of a duplicate of #396 .  The gist of it is that it won't really work how you expect. It won't catch certain things you'd want it to catch, and will catch things you wouldn't want it to catch.

IvanDimanov:
That's one big and very helpful gist, thak you :)
I guess promises really got it this time.
Hope in future there still will be solutions that uses async

PR #712, opened 2015-02-05
Added settle function

coffutt:
Added a 'settle' function that will allow all tasks to finish executing (in parallel) and then callback to the user. 

The goal of this function is to solve for when the user wants to execute tasks in parallel and wants them all to finish executing and return errors or results so that the handler can handle errors on a per task basis AND know what the return of the successful tasks was. With parallel, I can get an error and the results of any tasks finished at the time of error, but I'm unsure of the state of any tasks not complete at time of error.

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,6 +141,7 @@ Usage:
 * [`series`](#seriestasks-callback)
 * [`parallel`](#parallel)
 * [`parallelLimit`](#parallellimittasks-limit-callback)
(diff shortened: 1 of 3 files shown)


aearly:
This is an often requested feature, :+1: .  One nit would be to make the results property returned an array, rather than just the second arg -- some async functions callback with multiple values.

aearly:
Duplicate of #531

PR #727, opened 2015-03-05
Adding an error function called when a worker encounters an error

Antonhansel:
I'm using async.queue to recursively crawl folders on a distant API. 
The workers can push new tasks to the queue.

For my use case, the simplest and cleanest way to handle/log errors without stoping the async execution (When the API fails to respond for example) is  to call an error handling function set up the same way of drain();

````
q.taskQueue.error = function(err) {
      console.log(err);
  };
````

You can then handle the error the way you wish.

### DIFF
--- a/lib/async.js
+++ b/lib/async.js
@@ -769,10 +769,12 @@
         var workers = 0;
         var q = {
             tasks: [],
+            errors: [],
             concurrency: concurrency,
             saturated: null,
             empty: null,
             drain: null,
+            error: null,
             started: false,
             paused: false,
             push: function (data, callback) {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by deltaidea** [comment]
:+1: for the idea. This is useful.

1. Should add a test for this.
2. Should squash into a single commit.

**Comment by deltaidea** [comment]
For now, an alternative looks like this:

```CoffeeScript
onError = ( error ) ->
  console.log error.message

q = async.queue ( task, done ) ->
  doSomethingAsync task, ( error ) ->
    if error?
      onError error
      # optionally, call q.kill() here.
```

Pardon my CoffeeScript.

aearly:
The queue implementation has changed since this PR -- you'll have to reimplement this.  You can submit a new PR with docs and tests.

PR #734, opened 2015-03-15
Component.json: Add "main" entry

marcelklehr:


aearly:
Fixed in bb330627

marcelklehr:
Thank you!

PR #735, opened 2015-03-18
only call callback once, even if the callback errors

dughibot:


aearly:
Can you provide a test case for what this guards against?

aearly:
These were all fixed as a side effects of other changes.

PR #740, opened 2015-03-30
Allow cargo payload to be changed on the fly

gmaclennan:
According to [the docs](https://github.com/caolan/async/blob/master/README.md#cargo) you should be able to change the `payload` property of a `cargo` object on the fly.


aearly:
Closing in favor of #744.

PR #744, opened 2015-04-09
Making the change of payload size for cargo on runtime work

xalakox:
I believe https://github.com/caolan/async/pull/740 is incomplete since you are only doing it in the push process and does not applies to runtime for the pending jobs.


aearly:
The cargo logic has changed significantly -- can you re-do this work?  Also I need a test for this behavior.

PR #750, opened 2015-05-04
callbackNextTick method added to API

mstdokumaci:
Sometimes an async function needs tor return  a sync result. For the sake of consistency of async flow; we need to call the callback function wrapped in a nextTick block. This new method simplifys this effort.

Example:
```js
function asyncFn (other_arguments, callback) {
  if (cached_result) {
    return async.callbackNextTick(callback, cached_result);
  }
  some_async_call(some_arguments, callback);
}
```

### DIFF
--- a/lib/async.js
+++ b/lib/async.js
@@ -656,6 +656,13 @@
         };
     };
 
+    async.callbackNextTick = function (fn) {
+        var args = Array.prototype.slice.call(arguments, 1);
+        return async.nextTick(function () {
+            return fn.apply(null, args);
+        });
+    };
+
     var _concat = function (eachfn, arr, fn, callback) {
         var r = [];
         eachfn(arr, function (x, cb) {


aearly:
So this is like `async.nextTick`, except you can pass additional parameters to the function?

What about augmenting `async.nextTick` and `async.setImmediate` to do that?

aearly:
Ah, this type of behavior was voted down in iojs:  https://github.com/iojs/io.js/pull/1077

You can also accomplish the  same thing by doing:

```js
async.nextTick(callback.bind(null, cached_result));
```

aearly:
Well, looks like this was actually added:  https://github.com/nodejs/io.js/commit/10e31ba56c676bdcad39ccad22ea9117733b8eb5

aearly:
Closing, we'll track this in #940.

PR #751, opened 2015-05-07
added blocking parameter to priorityQueue

lmarcon:
I needed a priorityQuery with the ability to
1. process tasks concurrently
2. wait before processing lower priority tasks until those with higher priority are over, even with free workers 

Real (trivial) use case: in my list of tasks, I have an higher priority task (foo) that lasts longer than a lower priority task (bar), the concurrency is set to 2. I'd like to have them executed in the same priority order, so that bar doesn't start until foo is completed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1137,13 +1137,36 @@ q.unshift({name: 'bar'}, function (err) {
 ---------------------------------------
 
 <a name="priorityQueue" />
-### priorityQueue(worker, concurrency)
+### priorityQueue(worker, concurrency, blocking)
 
-The same as [`queue`](#queue) only tasks are assigned a priority and completed in ascending priority order. There are two differences between `queue` and `priorityQueue` objects:
+The same as [`queue`](#queue) only tasks are assigned a priority and completed in ascending priority order. If `concurrency > 1` and `blocking` is `true`, the process will wait until all the tasks with the current priority are over before starting tasks with a lower priority (higher value), even with available workers.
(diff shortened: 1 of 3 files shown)


aearly:
In the next version of async, you should be able to change the concurrency reliably on the fly.  You could do something like:

```js
if (foo.highPriority) {
  originalConcurrency = q.concurrency;
  q.concurrency = 1;
  q.push(foo, Inifnity, function () {
     q.concurrency = originalConcurrency;
  });
}
q.push(bar);
```

Would that be sufficient?  I'm kind of against this feature due to the complexity it introduces.

lmarcon:
That's fine for my really trivial example. What if I have several tasks, let's say:

    task1a, priority 1
    task1b, priority 1
    task2a, priority 2
    task2b, priority 2
    task2c, priority 2
    task2d, priority 2
    ... // lower priority tasks

and I want the concurrency to be 3 (or more), but still need to wait until tasks of priority 1 are over? And what if I have several of these queues in my code (or the queue is dynamically generated)?

I still believe it could be useful sometimes, but If you think the complexity introduced is too much to justify the merge, I understand.

In this last example, I'd need to use the code you provided in all queues instances, which could be fine as long as it's only me with this need - maybe I just need to redesign my code a bit.

aearly:
I feel this is too complicated to add to `queue`.  I'd recommend using 2 `priorityQueues`, and `.pause()`ing one when a blocking task is needed.

PR #759, opened 2015-05-20
perf(slice): fix not to use `Array#slice`

suguru03:
I fixed not to use `Array.prototype.slice.call(arguments)`.
This `_baseSlice` is based on [lodash](https://github.com/lodash/lodash/blob/master/lodash.src.js#L2711-L2731).

- [improve performance](https://github.com/suguru03/neo-async/commit/be3997f2c94baa82b1f839c4e25a0c60c5103f91)
- [Optimiztion-killers#leaking-arguments](https://github.com/petkaantonov/bluebird/wiki/Optimization-killers#32-leaking-arguments)

### DIFF
--- a/lib/async.js
+++ b/lib/async.js
@@ -97,6 +97,23 @@
         return keys;
     };
 
+    var _baseSlice = function (arr, start) {
+        start = start || 0;
+        var index = -1;
+        var length = arr.length;
+
+        if (start) {
+          length -= start;
+          length = length < 0 ? 0 : length;
+        }
+        var result = Array(length);
+
+        while (++index < length) {
+          result[index] = arr[index + start];
+        }
+        return result;
+    };
+
     //// exported async module functions ////
 
     //// nextTick implementation with browser-compatible fallback ////
@@ -338,19 +355,19 @@
 
     var doParallel = function (fn) {
         return function () {
-            var args = Array.prototype.slice.call(arguments);
+            var args = _baseSlice(arguments);
             return fn.apply(null, [async.each].concat(args));
         };
     };
     var doParallelLimit = function(limit, fn) {
         return function () {
-            var args = Array.prototype.slice.call(arguments);
+            var args = _baseSlice(arguments);
             return fn.apply(null, [_eachLimit(limit)].concat(args));
         };
     };
     var doSeries = function (fn) {
         return function () {
-            var args = Array.prototype.slice.call(arguments);
+            var args = _baseSlice(arguments);
             return fn.apply(null, [async.eachSeries].concat(args));
         };
     };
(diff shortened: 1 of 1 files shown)


aearly:
Hmnn, I merged this, but based on the benchmarks it's not significantly faster.  It doesn't hurt, though.

ronkorving:
You're still leaking arguments into baseSlice, right? How does this prevent deoptimization?

aearly:
It may not prevent full deoptimization, but after I fixed the benchmarks, this change is about 30% faster for functions like `waterfall`.

```
$ ./perf/benchmark.js -g waterfall 0.9.2 1.0.0
Comparing 0.9.2 with 1.0.0
waterfall(10) 0.9.2 x 2,444 ops/sec ±8.24% (23 runs sampled), 0.4ms per sample
waterfall(10) 1.0.0 x 3,420 ops/sec ±6.63% (25 runs sampled), 0.3ms per sample
1.0.0 is faster
waterfall(100) 0.9.2 x 317 ops/sec ±6.33% (24 runs sampled), 3.2ms per sample
waterfall(100) 1.0.0 x 408 ops/sec ±6.45% (26 runs sampled), 2.5ms per sample
1.0.0 is faster
waterfall(1000) 0.9.2 x 33.19 ops/sec ±6.45% (21 runs sampled), 30.1ms per sample
waterfall(1000) 1.0.0 x 44.19 ops/sec ±5.76% (25 runs sampled), 22.6ms per sample
1.0.0 is faster
1.0.0 faster overall (25ms total vs. 34ms total)
1.0.0 won more benchmarks (3 vs. 0)
```

ronkorving:
Alright, cool :) I wonder why, but I guess the numbers don't lie.

aearly:
I think passing `arguments` wholesale to another function isn't as bad as any of the other `arguments` related performance killers.  It's also possible V8 has gotten better in the 2 years since that performance killers article was written.

I think the fastest way still will be an in-line slice, like that article reccomends. but I wouldn't want to do that without some sort of macro.  ES6 + Babel also might be the way to go, with the rest and spread operators:

```js
function (err, ...args) {
  if (err) {
    return mainCallback(err);
  }
  nextCallback(null, ...args)
}
```

This is a pattern we do all over the place in async (or something similar), and with babel it gets compiled down to an inline slice and an `apply()` at the end, very similar to what we do manually with `_baseSlice`.

ronkorving:
Wouldn't it be nice if the V8 team kept a human readable list (and kept it up-to-date) of deoptimize triggers? :) Honestly, with a library like async, just doing it inline *may* be worth it, but I'll leave that to the authors.

PR #763, opened 2015-05-21
fix(benchmark): fix to enable to use second argument version

suguru03:
This PR enable to take correct benchmark. :smile:
Please review it.


aearly:
Ah thanks, I knew things were suspiciously even!

suguru03:
thanks!

PR #764, opened 2015-05-21
async.mySeries

lutaoact:
```
async.series({
    one: function(callback){
        setTimeout(function(){
            callback(null, 1);
        }, 200);
    },
    two: function(callback, results){
        console.log(results.one); //we can get "1" here
        setTimeout(function(){
            callback(null, 2);
        }, 100);
    }
},
function(err, results) {
    // results is now equal to: {one: 1, two: 2}
});
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,12 +1,36 @@
 # Async.js
 
+## extension
+```js
+myasync = require('myasync');
+
+myasync.mySeries({
+  one: function(next) {
+    next(null, 'one');
+  },
+  two: function(next, res) {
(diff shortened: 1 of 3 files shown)


aearly:
You can already do this with the existing `async.series`.

lutaoact:
@aearly I want to get the results not only in the final callback, but also in the task list. It means I can access `results.one` in the function `two`.

aearly:
That kind of behavior is already covered by `async.auto` and `async.waterfall`.

PR #769, opened 2015-05-25
ensureAsync

aearly:
Adds a new utility function for wrapping async functions.  It ensures that the function always calls is callback on a later tick in the event loop.  I'm also considering using this function internally to prevent stack overflows with synchronous iteration functions.


aearly:
I feel like there's a better name for this function, but going to use `ensureAsync` for now.

Issue #747, opened 2015-04-17
async.queue and concurrency


fprivitera:
if I change onthefly  the q.concurrency (of var q=async.queue(... ) ) there is no change happening.

are there any other workarounds?

thanks

aearly:
Is this related to #512 ?  As far as I know, there's no clean way to change the concurrency on the fly.

fprivitera:
no,
it is different : If I create a queue q with concurrency 5 and later I
change to q.concurrency=10, q.running() is still <=5.



On Sat, Apr 18, 2015 at 9:23 PM, Alexander Early <notifications@github.com>
wrote:

> Is this related to #512 <https://github.com/caolan/async/pull/512> ? As
> far as I know, there's no clean way to change the concurrency on the fly.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/caolan/async/issues/747#issuecomment-94193828>.
>



-- 
__________________________________________________________


*Filippo Privitera *|  Head of Technology


Milan | Rome | New York | Shanghai

Beintoo Spa - Corso di Porta Romana, 68 - 20122 Milano - Italy - Office
(+39) 02.97.687.959

This email is reserved exclusively for sending and receiving messages
inherent working activities, and is not intended nor authorized for
personal use. Therefore, any outgoing messages or incoming response
messages will be treated as company messages and will be subject to the
corporate IT policy and may possibly to be read by persons other than by
the subscriber of the box. Confidential information may be contained in
this message. If you are not the address indicated in this message, please
do not copy or deliver this message to anyone. In such case, you should
notify the sender immediately and delete the original message.

fprivitera:
why we don't add a function on async.queue like this?

setConcurrency: function(c) {
                q.concurrency=c
                if (!q.paused && workers < q.concurrency && q.tasks.length) {
                    var end=q.concurrency-workers
                    for (var i =0;i<end;i++) {
                        q.process()
                    }
                }

            }

justincy:
@fprivitera is half right. The concurrency is only updated when new items are pushed onto the queue. That behavior can be fixed without adding the proposed `setConcurrency` function.

justincy:
> The concurrency is only updated when new items are pushed onto the queue.

I need to revise that statement to say "The concurrency can only be updated on the same event loop in which an item was added to the queue. And it can only be decreased, not increased."

I just added a failing test case. There was previously a test case for changing concurrency but it didn't correctly test situations where the concurrency can't be changed.

[...]

PR #772, opened 2015-06-01
Queue concurrency

justincy:
Fix for #747

### DIFF
--- a/lib/async.js
+++ b/lib/async.js
@@ -886,7 +886,7 @@
                         }
                         workers += 1;
                         var cb = only_once(_next(q, tasks));
-                        async.setImmediate(worker(data, cb));
+                        worker(data, cb);
                     }
                 }
             },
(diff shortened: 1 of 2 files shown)


justincy:
Worked in node v0.10 but failed in v0.12 and iojs. I'll investigate.

lib/async.js
```diff
@@ -834,8 +834,21 @@
                 if (q.tasks.length === q.concurrency) {
                     q.saturated();
                 }
-                async.setImmediate(q.process);
             });
+            setImmediate(q.process);
```
aearly: Use async.setImmediate here, so it falls back to other methods when it's not available in the environment.

lib/async.js
```diff
@@ -859,32 +872,22 @@
...
+                        var data = _map(tasks, function (task) {
+                            return task.data;
+                        });
+    
+                        if (q.tasks.length === 0) {
+                            q.empty();
+                        }
+                        workers += 1;
+                        var cb = only_once(_next(q, tasks));
+                        setImmediate(worker(data, cb));
```
aearly: Likewise, `async.setImmediate`.

justincy:
@aearly Fixed.

Issue #635, opened 2014-09-30
Limit number of concurrent tasks in `async.auto`


vergenzt:
I'm using `async.auto` to spawn some subprocesses for tasks in a dependency tree. Right now it just runs as many tasks as it can run at a time, but I'd like to add an option to limit the total amount of tasks running at a given time, in the same way that `async.queue` currently works.

cryptoquick:
👍

PR #779, opened 2015-06-06
Implement contextSeries - alternative to waterfall

macobo:
This function is mainly useful as an alternative to watefall in versions of javascript, which support destructing.

Main problem with waterfall is that it becomes hard to debug/refactor after initial write due to needing to passaround arguments in _exactly_ the correct order. This function when used with destructing gets around that problem by allowing to name steps in the series and then each function can grab results it cares about.

Example (in coffeescript):
```coffeescript
# Adapted from one of my work scripts

getCurrentShards = (r, callback) -> # ...
cleanupOldShardMoves = ({shardInfo}, callback) -> # ...
createTask = ({shardInfo}, callback) -> # ...
setMoveStatus = (newStatus, {taskId, state}, callback) -> # ...
lockShards = ({taskId, shardInfo}, callback) -> # ...
# etc...

tasks = [
  ['shardInfo', getCurrentShards]
  cleanupOldShardMoves
  ['taskId', createTask]
  ['state', _.partial(setMoveStatus, 'initial')]
  lockShards
  _.partial(setMoveStatus, 'moving')
  moveInBatches
  _.partial(setMoveStatus, 'unlocking')
  unlockShards
  ['timestamp', updateMetadata]
]
async.contextSeries tasks, (err, {timestamp, taskId}) ->
  # ...
  log.info "Shard moves completed successfully at #{format timestamp} for task #{taskId}"

```

This function can prove to be even more useful by adding an optional `seedData` object with values which are made available to each task in the contextSeries array. I have omitted that however due to it not really fitting in with the rest of the API.

### DIFF
--- a/lib/async.js
+++ b/lib/async.js
@@ -661,6 +661,35 @@
         wrapIterator(async.iterator(tasks))();
     };
 
+    async.contextSeries = function(tasks, callback) {
+        callback = _once(callback || noop);
+        if (!_isArray(tasks)) {
+            return callback(new Error('First argument to contextSeries must be an array'));
+        }
+
+        var currentDependencies = [];
+        var autoGraph = {};
+        _arrayEach(tasks, function(task, index) {
+            var taskName, taskFn;
+            if (_isArray(task)) {
+                taskName = task[0];
+                taskFn = task[1];
+            } else {
+                taskName = "$"+index;
+                taskFn = task;
+            }
+
+            autoGraph[taskName] = function(callback, currentResults) {
+                taskFn(currentResults, callback);
+            };
+            if (currentDependencies.length) {
+                autoGraph[taskName] = currentDependencies.concat(autoGraph[taskName]);
+            }
+            currentDependencies.push(taskName);
+        });
+        async.auto(autoGraph, callback);
+    };
+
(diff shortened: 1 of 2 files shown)


aearly:
I don't feel like this is any more useful than `auto`.  Sure, you have guaranteed in-order execution, but you can easily do that in `auto` by setting your dependencies correctly.

aearly:
Also, the proposed feature in #635 -- essentially `autoLimit` -- might also be a way to enforce sequentiality.

macobo:
I would argue it is more useful than auto because of `auto`s downright weird argument ordering which kills function composability.

As mentioned before, we're using it as an alternative to `waterfall` which *always* is a huge pain to refactor due to having to pass arguments through in the exact correct order.

aearly:
I hear you about `auto`'s weird parameter ordering.  I want to swap the ordering, but need to figure out the best way to phase it in.

In the meantime, I often write a simple combinator that flips the arguments:

```js
function flip(fn) {
  return function (results, cb) {
    return fn(cb, results);
  };
}

function normalAsyncFunction(options, callback) {...}

async.auto({
  //..
  "task": ["dependency", flip(normalAsyncFunction)]
  //..
})
```

PR #790, opened 2015-06-14
Fix timesSeries documentation

perrin4869:
I'm not entirely sure what the original intention was so the fix should be reviewed first... but I think that this is what was meant


aearly:
Thanks!

PR #792, opened 2015-06-16
add interval parameter to async.retry - to delay the next attempt

edeustace:
I propose adding an optional `interval` parameter for `async.retry`, that will introduce a delay between attempts. I have a scenario where I want to retry a call, but not straight away.

This is a non-breaking change.


edeustace:
Closing - going to move interval + times to an option object that can be the 1st param.

PR #793, opened 2015-06-17
Feature/retry add optional interval

edeustace:
I propose adding an optional `interval` parameter for async.retry, that will introduce a delay between attempts. I have a scenario where I want to retry a call, but not straight away.

This is a non-breaking change.

`interval` may be passed in as part of an options object as the first parameter to `async.retry`: 

```javascript
    async.retry({times: 2, interval: 100}, foo, bar);
````
The existing function signature remains in place: 

```javascript
    async.retry(2, foo, bar); // same as {times: 2, interval: 0}
````

### DIFF
--- a/README.md
+++ b/README.md
@@ -1459,7 +1459,7 @@ new tasks much easier (and the code more readable).
 ---------------------------------------
 
 <a name="retry" />
-### retry([times = 5], task, [callback, interval = 0])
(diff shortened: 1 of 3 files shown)


aearly:
Cool, a nice little enhancement.  I also prefer this API to your previous PR.  Good work! :100:

Issue #602, opened 2014-08-26
Whilst async truth test


gobwas:
Why in `whilst` method (and brothers) truth test designed sycnchronously? For me its very often reason to write my own whilst, with async test. May be we can add it? If yes, i can implement it.

Sorry, if Im wrong.

Issue #156, opened 2012-07-31
whilst should allow an asynchronous test function


brianmaissy:
It would be useful (to me at least) if whilst could optionally accept an asynchronous test function. It would work by delegating to the normal whilst, just storing local state about the last outcome of the test. Something like this:

```
whilst: function(test, action, callback){
    if(test.length > 0) newWhilst(test, action, callback)
    else {
        ...
    }
},

newWhilst: function(test, action, callback){
    test(function(state){
       whilst(function(){
            return state;
        },
        function(callback){
            action(function(){
                test(function(result){
                    state = result;
                    callback();
                });
            });
        },
        callback);
    });
},
```

Is there a reason this feature isn't implemented? If not I'd be happy to add it and submit a pull request

PR #800, opened 2015-06-23
during & doDuring with async test

uzyn:
It's basically `whilst` and `doWhilst` with async truth test.

Addresses #156 #602 

Will add documentation if maintainer is okay with this PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -192,6 +192,8 @@ Usage:
 * [`doWhilst`](#doWhilst)
 * [`until`](#until)
 * [`doUntil`](#doUntil)
+* [`during`](#during)
+* [`doDuring`](#doDuring)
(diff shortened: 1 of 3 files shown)


aearly:
:+1:, since several people have asked for it.

Also -- I wonder if the test should be passed a callback of the form `function (err, truth) {...}` rather than just `function (truth) {...}`.  I realize that most of async`s predicate-accepting functions expect a callback with just a `truth` parameter, but it's something we're going to change in async 2.0.

uzyn:
Awesome. Let me update that and the documentation as well.

uzyn:
Alright. Updated as suggested.

Thanks @aearly

PR #802, opened 2015-06-24
Add expect statements to (many) tests

megawac:
Doing this made me question my sanity, but nonetheless, this should help to prevent some timeout errors in the future :)


aearly:
> question my sanity

Indeed  :) 

It definitely will help things along though -- things wall fail faster and prevent some of nodeunit's hard-to-interpret error messages.

PR #804, opened 2015-06-24
Improve the detection of global/root object

megawac:
Refer to https://github.com/jashkenas/underscore/issues/2152. We were using a similar approach to `async`s current one in underscore until https://github.com/jashkenas/underscore/pull/2153.

The two main benefits of this change are:
 - Forwards compatibility with `strict` mode and some tooling (see https://github.com/jashkenas/underscore/issues/2152)
 - `WebWorker` support, you can't use `window` in a `WebWorker` but you can use `self`

Aside: [here's the detection method from lodash](https://github.com/lodash/lodash/blob/master/lodash.js#L248-254)

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -20,6 +20,7 @@
     "browser": true,
     "node": true,
     "globals": {
+        "self": true,
         "define": true
     }
 }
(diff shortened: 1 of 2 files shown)


aearly:
Should we explicitly check for `window`?  If async.js gets wrapped in something that adds strict mode, `this` in the check will be `null`.

megawac:
Self covers that case

aearly:
OK, cool. :+1:

PR #805, opened 2015-06-24
Setup make publish script

megawac:
This adds a `make` task which:

- Tests and builds async
- Adds a commit bumping the package.json and tags the commit
- Adds the `dist/` to the published package (only on github - useful for CDNs) (only for releases)
- Pushes to the commit and tag
- Publishs to npm

Usage `make release-patch`

Reference: https://github.com/davidchambers/xyz


aearly:
Would be nice if it also ran `support/sync-package-managers.js`.  It also looks like `xyz`creates a tag of the form `v1.2.3` -- can it also do `1.2.3`?  IIRC Bower expects the latter.

megawac:
Semver supports both tag formats and yeah, I saw the syncpackagemanagers
and didn't add it for whatever reason. I'll do that when I get back

megawac:
BTW is there an irc,slack or gitter chanmel for this project?

aearly:
I just added a gitter room.  Currently it's freezing any browser I load it with, though....

megawac:
> It also looks like xyzcreates a tag of the form v1.2.3 -- can it also do 1.2.3? IIRC Bower expects the latter.

I have wayyy to intimate knowledge of semver cause of jsdelivr and [semvish](https://github.com/megawac/SemVish)
```js
var S = require('semver')
S.satisfies('v0.1.1', '0.1.1')
```
If bower doesn't support `v*` tags then it's a bug on their end

PR #813, opened 2015-06-29
Build distribution file to lib/async.min.js

megawac:


aearly:
Should we do 1.3.1 now, or save this for the next release?

megawac:
Either or, thanks!

PR #818, opened 2015-07-01
Replace slice with restParam

megawac:
This has been known to greatly simplify code (heavily used in underscore and lodash) and be a great optimization (slicing arguments is sloooooooooooow).

If interested, refer to these threads for documentation, benchmarks, arguments, naming, etc:-)
- https://github.com/jashkenas/underscore/pull/2140
- https://github.com/lodash/lodash/issues/1122
- https://lodash.com/docs#restParam

### DIFF
--- a/lib/async.js
+++ b/lib/async.js
@@ -137,21 +137,29 @@
         }
     }
 
-    function _baseSlice(arr, start) {
-        start = start || 0;
-        var index = -1;
-        var length = arr.length;
-
-        if (start) {
-            length -= start;
-            length = length < 0 ? 0 : length;
-        }
-        var result = Array(length);
-
-        while (++index < length) {
-            result[index] = arr[index + start];
-        }
-        return result;
+    // Similar to ES6's rest param (http://ariya.ofilabs.com/2013/03/es6-and-rest-parameter.html)
+    // This accumulates the arguments passed into an array, after a given index.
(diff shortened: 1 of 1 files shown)


lib/async.js
```diff
@@ -137,21 +137,29 @@
...
+                case 0: return func.call(this, rest);
+                case 1: return func.call(this, arguments[0], rest);
+                case 2: return func.call(this, arguments[0], arguments[1], rest);
+            }
+            var args = Array(startIndex + 1);
+            for (index = 0; index < startIndex; index++) {
+                args[index] = arguments[index];
+            }
+            args[startIndex] = rest;
+            return func.apply(this, args);
```
megawac: Line 156-161 is currently unused. Thoughts? I was thinking about just commenting it out as it handles cases not handled by the switch statement

aearly:
Interesting -- simplifies the code a bit.  Is this actually faster, though?  What do the benchmarks in `perf` say?

megawac:
The bench results are essentially the same in all benchmarks +- a couple
milliseconds which makes sense because this is a synch part of a async
test.

aearly:
Ok, that's fine then.  When we switched from `Array.prototype.slice` to `_baseSlice`, there was a measurable performance increase!  This gives us the same performance with cleaner code, so :+1:

PR #819, opened 2015-07-01
Fix tag detection methodology

megawac:


perf/benchmark.js
```diff
@@ -3,11 +3,13 @@
 var _ = require("lodash");
 var Benchmark = require("benchmark");
 var exec = require("child_process").exec;
+var execSync = require("child_process").execSync;
```
aearly: `execSync` is only node 0.12 and later, right?  I still run these in node 0.10 to compare.

megawac: Ah you're right!

aearly: Change this to use the async `exec` and I'll merge this in :+1:

PR #821, opened 2015-07-01
Implement series via _parallel

megawac:
This seems to have been missed in all the `_parallel` changes


aearly:
I guess `_parallel` is a bit of a misnomer now -- perhaps  something like `_runTasks`?

PR #823, opened 2015-07-01
Implement doWhilst via whilst

megawac:


aearly:
Hmnn, this means that the second and later invocations of the `test` in `whilst` will be called with the callback arguments -- might lead to subtle breaking changes.  I'm sure most people are passing a function expression with no parameters, but some people might be passing a variadic method somehow.

megawac:
I mean, if people don't expect parameters they won't be handling them. I would be extremely surprised if this broke anyones code (plus its a nice little enhancement to `whilst` (same goes for `during` in the other pr)

aearly:
Okay, I guess we'll try it and see...

PR #827, opened 2015-07-02
Optimize applyEach* via creator function

megawac:
This is both an optimization (avoid restParam + concat) and a code simplification :D


aearly:
Man, its amazing how much _restParam simplifies... everything...

PR #829, opened 2015-07-02
Implement _detect via tester

megawac:
As I mentioned in the #828 pull request, this is easy to extend to `_detect`. Also should `detectLimit` be added?


aearly:
`detectLimit` seems trivial to add -- most of the work would be covering it with tests.

PR #834, opened 2015-07-03
Minor auto simplifications

megawac:


lib/async.js
```diff
@@ -571,7 +574,7 @@
                 if (!(dep = tasks[requires[len]])) {
                     throw new Error('Has inexistant dependency');
                 }
-                if (_isArray(dep) && !!~dep.indexOf(k)) {
```
megawac: Note, `Array.prototype.indexOf` is ES5 while it seems async is targeting ES3?

aearly: Yep, It seems that way.  Maybe we can drop ES3 support in 2.0.  This was out in the wild for a good while without any bug reports.

aearly:
:+1:

PR #837, opened 2015-07-03
Implement style guide check via jscs

megawac:
Wheewh :weary: I need :coffee: 

Stricter rules (e.g. google or airbnb style guide) should probably follow

### DIFF
--- a/.jscsrc
+++ b/.jscsrc
@@ -1,0 +1,3 @@
+{
+  "validateIndentation": 4
+}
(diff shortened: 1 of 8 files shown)


aearly:
I'm :+1: for standardizing on indentation, but :-1: for the other stuff.  I don't really care about whether `function` should have a following space or single vs. double quotes.  Can you make the `.jscsrc` unopinionated about that kind of stuff?

megawac:
You can, I just had it generate this change set using `google`s style guide preset rules and then disabled the presets. I do think its good to follow a consistent style for most of this stuff, especially in small, heavily used code bases.

megawac:
Btw it's `jscs <file> --fix` to auto correct the formatting

megawac:
I'd still like to enforce slightly stricter style in the future

aearly:
What value does a stricter style bring?

PR #840, opened 2015-07-08
asyncify promisified functions

zartdinov:
Most of Promise realisations doesn't contain `nodeify` method. Here is extention of async wrapper for promisified functions.
For example promise-styled [openpgp](https://github.com/openpgpjs/openpgpjs) library functions looks like
```javascript
openpgp.key.generate(options).then(function(key) {
    // success
}).catch(function(error) {
    // failure
});
```
And can be easily wrapped to use with favourite async:)
```javascript
async.asyncify(openpgp.key.generate);
```

### DIFF
--- a/lib/async.js
+++ b/lib/async.js
@@ -1186,10 +1186,18 @@
             var result;
             try {
                 result = func.apply(this, args);
+                // if result is Promise object
+                if (typeof result.then === "function") {
+                    result.then(function(values) {
+                        var args = [null].concat(values);
+                        callback.apply(this, args);
+                    }).catch(callback);
+                } else {
+                    callback(null, result);
+                }
             } catch (e) {
                 return callback(e);
             }
-            callback(null, result);
         });
     };
 
--- a/lib/async.js
+++ b/lib/async.js
@@ -1186,17 +1186,18 @@
             var result;
             try {
                 result = func.apply(this, args);
-                // if result is Promise object
-                if (typeof result.then === "function") {
-                    result.then(function(values) {
-                        var args = [null].concat(values);
-                        callback.apply(this, args);
-                    }).catch(callback);
-                } else {
-                    callback(null, result);
-                }
             } catch (e) {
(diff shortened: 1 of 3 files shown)


lib/async.js
```diff
@@ -1186,10 +1186,18 @@
             var result;
             try {
                 result = func.apply(this, args);
+                // if result is Promise object
+                if (typeof result.then === "function") {
+                    result.then(function(values) {
+                        var args = [null].concat(values);
+                        callback.apply(this, args);
```
megawac: There can be only one value from promise so can you do 
```js
callback.call(null, value)
```

megawac:
:+1: can you add some unit tests (feel free to add a dev dependency for some promise implementation, e.g. NPO/rsvp/bluebird/es6-promise)

lib/async.js
```diff
@@ -1186,10 +1186,18 @@
...
+                    }).catch(callback);
+                } else {
+                    callback(null, result);
```
aearly: We don't want to have the callback call in the try/catch block.

aearly: Yep: The failing test is the one that guards against this: https://travis-ci.org/caolan/async/jobs/70115957

zartdinov:
Done!

test/test-async.js
```diff
@@ -4267,5 +4267,49 @@ exports['asyncify'] = {
...
-    }
+    },
+
+    'promisified': [
+        'native-promise-only',
+        'bluebird',
+        'es6-promise',
+        'rsvp'
+    ].reduce(function(promises, name) {
+        var Promise = require(name);
```
aearly: You might want to make these tests node-only, because this won't work in the browser.

zartdinov:
I think node-only for a while, but all promise libraries differently support browsers.

aearly:
It's not the promise implementation, but the `require(name)` statement that won't work.  We can fix the tests for browsers later.

PR #844, opened 2015-07-11
Stricter type checking

megawac:


lib/async.js
```diff
@@ -1186,7 +1192,7 @@
                 return callback(e);
             }
             // if result is Promise object
-            if (typeof result !== 'undefined' && typeof result.then === "function") {
+            if (_isObject(result) && typeof result.then === "function") {
```
megawac: Could also be loser here and do `result != null` currently this will error if result is null tho

aearly:
Overlooked this.  Looks good to me.

PR #854, opened 2015-07-20
Drop unused restParam case

megawac:


aearly:
:+1:  A bit weird to have a switch without a default, though...

PR #858, opened 2015-07-21
Added detection for using in  Atom/Electron

xiamx:
Atom/Electron also exposes module object to the renderer Javascript
context. Currently async.js detects Electron as NodeJs, where as it
should be detected as browser, as suggested by
https://github.com/atom/electron-starter#why-does-my_favorite_library-not-work--do-weird-stuff

The detection snippet is from
https://github.com/atom/electron/issues/2288

### DIFF
--- a/lib/async.js
+++ b/lib/async.js
@@ -1203,8 +1203,12 @@
         });
     };
 
+    // Electron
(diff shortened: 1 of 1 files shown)


megawac:
See https://github.com/atom/electron/issues/254, instead of loading `async` globally load it via require. I think merging this would break more epoples code than help users.

PR #866, opened 2015-07-28
Add detectLimit

markyen:
This function is already in the documentation but was not in the implementation.


aearly:
:+1:  good catch

megawac:
Probably worth a patch release

Also yeah, I think it was going to get added in #847 :p

aearly:
Ah, forgot about that.  This can fill in until we're ready for that PR.

megawac:
Definitely, and adds tests. Thanks @markyen!

Issue #864, opened 2015-07-28
"ReferenceError: obj is not defined" when using async.each


yourfriendgerasim:
I guess this is just a small typo, but happens to me, when I use the current async.js file. Minimal working example:

```html
<!DOCTYPE html>
<html>
    <script src="async.js"></script>
    <script type="text/javascript">
        async.each([1,2,3,4] , function iterator(item, callback){
              callback();
        }, function (err) {
              if (err) {console.log(err);}
        });
    </script>
</html>

aearly:
Seems to work for me:  http://codepen.io/anon/pen/MwzpeY

PR #867, opened 2015-07-29
Fix `ReferenceError` of `eachOf` and remove unused function

suguru03:
Related to https://github.com/caolan/async/issues/864, https://github.com/caolan/async/pull/853

This `obj` is not defined on dist.
But, I think [this improvement](https://github.com/caolan/async/pull/853) needs to applied to `lib'.

### DIFF
--- a/dist/async.js
+++ b/dist/async.js
@@ -71,12 +71,6 @@
         );
     }
 
-    function _each(coll, iterator) {
-        return _isArrayLike(coll) ?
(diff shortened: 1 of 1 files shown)


megawac:
Hi @suguru03 mind doing this on `lib/` (and picking #853 over)

suguru03:
I tried to fix this to `lib`, but some test cases are called synchronously.
So, `--running` is called synchronously and this implementation does not pass.

PR #872, opened 2015-07-31
Revert "Implement eachOf via keyIterator"

megawac:
Reverts caolan/async#853


dist/async.js
```diff
@@ -224,22 +224,22 @@
     async.forEachOf =
```
aearly: Are lib/ and dist/ in sync here?

PR #873, opened 2015-08-01
Allow map to return objects

megawac:
Map currently will attempt to assign keys to arrays and because of a bug in `nodeunit` this passes unit tests. I think map should either always return an array (ala lodash/underscore) for objects or do as this PR does

/cc @aearly

### DIFF
--- a/lib/async.js
+++ b/lib/async.js
@@ -353,7 +353,8 @@
 
     function _asyncMap(eachfn, arr, iterator, callback) {
         callback = _once(callback || noop);
-        var results = [];
+        arr = arr || [];
+        var results = _isArrayLike(arr) ? [] : {};
         eachfn(arr, function (value, index, callback) {
             iterator(value, function (err, v) {
                 results[index] = v;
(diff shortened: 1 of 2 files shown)


aearly:
I'm not sure what the "correct" thing to do here is.  The way it works now is just a happy accident of using `eachOf` internally.  I agree that `map` should always return arrays.  Lodash returns an array when you pass an array to `_.map`, and has a `_.mapValues` function that returns an object.  We could add `async.mapValues` (and `mapValuesSeries` and `mapValuesLimit`), but as you can see, it gets out of hand.

Once we support iterators, one thing that will be really useful is being able to have `map` callback with another iterator -- it will enable doing some cool stuff with lazy sequences.  But perhaps we need another method for a mapping operation that keeps the type of the iterable you pass in.

aearly:
Another thought -- Haskell and other FP languages have the concept of an `fmap` function -- a function that takes a functor (basically anything that is mappable) and returns a new instance of that functor with the mapping function applied to each element.  Perhaps we could call the new method `async.fmap`?  (and `fmapSeries` and `fmapLimit`)

aearly:
Going to merge this for now, we can have map return an array later.

PR #891, opened 2015-08-23
Added workersList() method into queue

kolomiichenko:
`workersList()` - a function returning the array of items currently being processed. This can be useful when it is not need push the new job to the queue if it's in process.


sadovnychyi:
[LGTM](http://lgtm.in/i/qJOa3EHRp)

[:+1:](http://lgtm.in/u/qJOa3EHRp)

aearly:
Why do you need this?  What problem does it solve?

kolomiichenko:
@aearly, in my situation, every task working for a few seconds(minutes), so I need to display progress queue. But in queue.tasks object I can't see tasks currently being processed. This code will solve this problem. Secondly, his can be useful when it is not need push the new job to the queue if it's in process.

lib/async.js
```diff
@@ -863,8 +863,17 @@
         function _next(q, tasks) {
             return function(){
                 workers -= 1;
+
```
megawac: spliced?

kolomiichenko: @megawac, I do not know what I guided when giving named. Renamed to `removed`. :smile:

megawac:
@aearly @kolomiichenko is this actually useful? I feel like you can get the same information on the progress by inspecting `queue.length`, `queue.running` and `queue.concurrency`

I want to better understand this use case, as this has become a bottleneck for queue

aearly:
I feel like the above comment explains it pretty well.  Sometimes it is useful to know exactly which objects are being processed by the queue.

PR #898, opened 2015-08-31
Adds support for es6 iterators

### LINKED ISSUE #839: Support ES6's Iterators

megawac:
Still not sure if this is the right take to address this. The main issues is I expect there will be a perf hit as a result of creating a new array (`[value, key]`) at each step iteration step, instead of  just passing keys directly (benchmarks to follow).

Fixes #839 

**Benchmarks**
```
Comparing v1.3.0 with current on Node v2.3.3
each(10) v1.3.0 x 36,160 ops/sec ±3.60% (24 runs sampled), 0.0277ms per run
each(10) current x 35,297 ops/sec ±1.45% (28 runs sampled), 0.0283ms per run
v1.3.0 is faster
each(300) v1.3.0 x 6,017 ops/sec ±1.55% (26 runs sampled), 0.166ms per run
each(300) current x 5,883 ops/sec ±1.06% (28 runs sampled), 0.17ms per run
v1.3.0 is faster
each(10000) v1.3.0 x 215 ops/sec ±1.33% (28 runs sampled), 4.64ms per run
each(10000) current x 207 ops/sec ±3.90% (28 runs sampled), 4.84ms per run
v1.3.0 is faster
eachSeries(10) v1.3.0 x 19,552 ops/sec ±1.27% (27 runs sampled), 0.0511ms per run
eachSeries(10) current x 18,455 ops/sec ±4.19% (26 runs sampled), 0.0542ms per run
v1.3.0 is faster
eachSeries(300) v1.3.0 x 1,115 ops/sec ±2.07% (27 runs sampled), 0.897ms per run
eachSeries(300) current x 1,023 ops/sec ±2.77% (25 runs sampled), 0.978ms per run
v1.3.0 is faster
eachSeries(10000) v1.3.0 x 35.27 ops/sec ±1.47% (29 runs sampled), 28.4ms per run
eachSeries(10000) current x 31.43 ops/sec ±3.30% (27 runs sampled), 31.8ms per run
v1.3.0 is faster
eachLimit(10) v1.3.0 x 25,702 ops/sec ±4.46% (28 runs sampled), 0.0389ms per run
eachLimit(10) current x 25,670 ops/sec ±2.65% (27 runs sampled), 0.039ms per run
Tie
eachLimit(300) v1.3.0 x 2,282 ops/sec ±2.68% (28 runs sampled), 0.438ms per run
eachLimit(300) current x 2,321 ops/sec ±3.81% (25 runs sampled), 0.431ms per run
Tie
map(10) v1.3.0 x 32,804 ops/sec ±3.23% (25 runs sampled), 0.0305ms per run
map(10) current x 32,166 ops/sec ±1.56% (27 runs sampled), 0.0311ms per run
Tie
map(300) v1.3.0 x 5,134 ops/sec ±1.12% (28 runs sampled), 0.195ms per run
map(300) current x 4,830 ops/sec ±2.39% (28 runs sampled), 0.207ms per run
v1.3.0 is faster
map(10000) v1.3.0 x 172 ops/sec ±2.41% (28 runs sampled), 5.81ms per run
map(10000) current x 165 ops/sec ±2.19% (28 runs sampled), 6.07ms per run
v1.3.0 is faster
mapSeries(10) v1.3.0 x 17,971 ops/sec ±1.44% (27 runs sampled), 0.0556ms per run
mapSeries(10) current x 17,425 ops/sec ±2.19% (27 runs sampled), 0.0574ms per run
v1.3.0 is faster
mapSeries(300) v1.3.0 x 969 ops/sec ±3.29% (25 runs sampled), 1.03ms per run
mapSeries(300) current x 962 ops/sec ±2.49% (27 runs sampled), 1.04ms per run
Tie
mapSeries(10000) v1.3.0 x 30.75 ops/sec ±4.29% (26 runs sampled), 32.5ms per run
mapSeries(10000) current x 30.44 ops/sec ±5.64% (26 runs sampled), 32.8ms per run
Tie
mapLimit(10) v1.3.0 x 24,858 ops/sec ±1.12% (27 runs sampled), 0.0402ms per run
mapLimit(10) current x 25,084 ops/sec ±1.20% (28 runs sampled), 0.0399ms per run
Tie
```

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -19,6 +19,7 @@
     "sub": true,
     "browser": true,
     "node": true,
+    "esnext": true,
     "globals": {
         "self": true,
         "define": true,
--- a/lib/async.js
+++ b/lib/async.js
@@ -77,12 +77,6 @@
         );
     }
 
-    function _each(coll, iterator) {
-        return _isArrayLike(coll) ?
-            _arrayEach(coll, iterator) :
-            _forEachOf(coll, iterator);
-    }
-
     function _arrayEach(arr, iterator) {
         var index = -1,
             length = arr.length;
@@ -230,23 +224,26 @@
     async.eachOf = function (object, iterator, callback) {
         callback = _once(callback || noop);
         object = object || [];
-        var size = _isArrayLike(object) ? object.length : _keys(object).length;
-        var completed = 0;
-        if (!size) {
-            return callback(null);
-        }
-        _each(object, function (value, key) {
+
+        var iter = _keyIterator(object);
+        var key, completed = 0;
+
+        while ((key = iter()) != null) {
+            completed += 1;
             iterator(object[key], key, only_once(done));
-        });
+        }
+
+        if (completed === 0) callback(null);
+
         function done(err) {
+            completed--;
             if (err) {
                 callback(err);
             }
-            else {
-                completed += 1;
-                if (completed >= size) {
-                    callback(null);
-                }
+            // Check key is null in case iterator isn't exhausted
+            // and done resolved synchronously.
+            else if (key === null && completed <= 0) {
+                callback(null);
             }
         }
     };
--- a/lib/async.js
+++ b/lib/async.js
@@ -19,6 +19,7 @@
         return !v;
     }
 
+
     // global on the server, window in the browser
     var previous_async;
 
@@ -32,6 +33,8 @@
     if (root != null) {
         previous_async = root.async;
     }
+
+    var iteratorSymbol = typeof Symbol == 'function' && Symbol.iterator;
 
     async.noConflict = function () {
         root.async = previous_async;
@@ -131,22 +134,32 @@
         return keys;
     };
 
-    function _keyIterator(coll) {
+    function _iterator(coll) {
         var i = -1;
         var len;
         var keys;
+
         if (_isArrayLike(coll)) {
             len = coll.length;
             return function next() {
                 i++;
-                return i < len ? i : null;
+                return i < len ? [coll[i], i] : null;
             };
-        } else {
+        }
+        else if (iteratorSymbol && coll[iteratorSymbol]) {
+            var iterator = coll[iteratorSymbol]();
+            return function next() {
+                var item = iterator.next();
+                if (item.done) return null;
+                return _isArray(item.value) ? item.value : [item.value];
+            };
+        }
+        else {
             keys = _keys(coll);
             len = keys.length;
             return function next() {
                 i++;
-                return i < len ? keys[i] : null;
+                return i < len ? [coll[keys[i]], keys[i]] : null;
             };
         }
     }
@@ -225,12 +238,12 @@
         callback = _once(callback || noop);
         object = object || [];
 
-        var iter = _keyIterator(object);
-        var key, completed = 0;
+        var iter = _iterator(object);
+        var item, completed = 0;
 
-        while ((key = iter()) != null) {
+        while ((item = iter()) != null) {
             completed += 1;
-            iterator(object[key], key, only_once(done));
+            iterator(item[0], item[1], only_once(done));
         }
 
         if (completed === 0) callback(null);
@@ -242,7 +255,7 @@
             }
             // Check key is null in case iterator isn't exhausted
             // and done resolved synchronously.
-            else if (key === null && completed <= 0) {
+            else if (item === null && completed <= 0) {
                 callback(null);
             }
         }
@@ -252,20 +265,20 @@
     async.eachOfSeries = function (obj, iterator, callback) {
         callback = _once(callback || noop);
         obj = obj || [];
-        var nextKey = _keyIterator(obj);
-        var key = nextKey();
+        var iter = _iterator(obj);
+        var item = iter();
         function iterate() {
             var sync = true;
-            if (key === null) {
+            if (item === null) {
                 return callback(null);
             }
-            iterator(obj[key], key, only_once(function (err) {
+            iterator(item[0], item[1], only_once(function (err) {
                 if (err) {
                     callback(err);
                 }
                 else {
-                    key = nextKey();
-                    if (key === null) {
+                    item = iter();
+                    if (item === null) {
                         return callback(null);
                     } else {
                         if (sync) {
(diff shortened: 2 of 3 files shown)


aearly:
:+1: The performance impact seems negligible.

megawac:
Probably should be tested in more environments before a decision is made
here

ex1st:
@megawac Maybe create an object {key, value} for a smaller penalty performance?
https://jsperf.com/new-array-vs-new-object-in-iterators

p.s. But not {0, 1}. This object will confuse V8 JIT optimizator.

megawac:
Afaik, arrays have historically been well optimized in v8

ex1st:
For iteration, but not for creation.

megawac:
Would you be interested in forking this branch and running the benchmarks?
It'd  be nice to squeeze every bit of juice out of this

ex1st:
I can do it, but little bit later (during this week).

_________________________________________________
Why "return _isArray(item.value) ? item : [item]" is bad idea:

```javascript
var mySet = new Set();
mySet.add("a");
mySet.add({});
mySet.add([1, 2]);

var  setIter = _iterator(mySet);

setIter(); // => ["a"];
setIter(); // => [{}];
setIter(); // => [1, 2], but we awaiting [[1, 2]];
```

megawac:
How would you suggest plucking keys from a map iterator in that case?

var m = new Map() m.set({a:1}, {b: 5}) m[Symbol.iterator]().next()
// Object {value: Array[2], done: false}

ex1st:
Best solution would be Array.from but not all environments supported this.

``` javascript
var m = new Map();
m.set({a:1}, {b:5});
m.set([1,2]);
....
else if (iteratorSymbol && coll[iteratorSymbol]) {
    var iterator = Array.from(coll)); // Array.from(coll) == > [ [{a:1}, {b:5}], [[1,2], undefined] ]
    len = iterator.length;
    return function next() {
       i++;
       return i < len ? [iterator[i][0], iterator[i][1]] : null;
     };
}
```

megawac:
Wouldn't that run into the same issue with sets of arrays and arrays of
arrays?

ex1st:
Other case of iterators return only value without key? Check instanceof Map before calls Array.from? :)

megawac:
Probably makes most sense to ignore *EntryIterator* types (such as the map
iterator) as typically the user is required to manually destructure the
arguments for these case (i.e. `for (let [x, y] of map){}`). Therefore we
can suggest our users use similar strategies when passing entry iterators
(`async.fn(iterator, function([val, key], next) {})`)

Thoughts @aearly @ex1st

ex1st:
You mean a users must manually convert such objects to array/collection?

aearly:
Is there a simple way to detect the difference between the Map style iterator (`[key, value]`) and the Array style iterator (`[value]`)?

megawac:
> You mean a users must manually convert such objects to array/collection?

Leave `value, key` iterators out of scope for the implementation and suggest how to handle it in documentation (i.e. `async.forEach( ([value, key], next) => {})`)

> Is there a simple way to detect the difference between the Map style iterator ([key, value]) and the Array style iterator ([value])?

I don't believe so

ex1st:
@megawac, sounds good. But still need some keys for forEachOf. Do you agree?
```javascript
else if (iteratorSymbol && coll[iteratorSymbol]) {
    var iterator = coll[iteratorSymbol]();
    return function next() {
      i++;
      var item = iterator.next();
      if (item.done) return null;
      return [item.value, i];
     };
}
```

```javascript
async.forEachOf( ([value, key], index, next) => {}))
```

@aearly, it's unreal.

megawac:
I would suggest always giving `undefined` for `async.*Of` functions when
iterating an iterators/generators

ex1st:
Ok :)

ex1st:
@megawac, async.map will return a empty array if keys will be 'undefined'

megawac:
Not if its implemented via _iterator

ex1st:
You suggest pass 'undefined' only in async.*Of, but _iterator must return keys always?

aearly:
One thing I discovered:  creating simple Objects is slightly faster than creating Arrays:  http://jsperf.com/object-vs-array-vs-class .  Most JS engines convert small, non-dynamic Objects to structs internally.   (I remember discussing this with @megawac  a while back).

This might make the cost of converting an interator that returns just values, to an iterator that returns {key, value} pairs.

jdalton:
From the benchmarks it looks like perf is ok. As for the arrays or objects I think arrays map better to a map which can be initialized as an array of key-value pairs. I'm not sure what you're wanting to provide though as I see if you're iterating an array you return `[value, index]` and if you're iterating an iterator you return `[value]`.

aearly:
Now that we're using lodash internally, would we get iterator support for free if lodash supported them?

jdalton:
> Now that we're using lodash internally, would we get iterator support for free if lodash supported them?

Lodash has limited iterator support. It currently only supports them in `_.toArray` and the chaining/sequence wrapper is itself and iterator and iterable.

aearly:
Now that we've majorly refactored Async, this should be revisited in another PR.

PR #900, opened 2015-09-04
Code changes to standardize style

rheh:
Trivial changes to style to standardize style of function definitions, if statement braces etc.  Style implemented follows that used in the majority of the code.

### DIFF
--- a/lib/async.js
+++ b/lib/async.js
@@ -39,7 +39,7 @@
     };
 
     function only_once(fn) {
-        return function() {
+        return function () {
             if (fn === null) throw new Error("Callback was already called.");
             fn.apply(this, arguments);
             fn = null;
@@ -47,7 +47,7 @@
     }
 
     function _once(fn) {
-        return function() {
+        return function () {
             if (fn === null) return;
             fn.apply(this, arguments);
             fn = null;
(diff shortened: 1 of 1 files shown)


aearly:
I'd prefer to remain un-opinionated about this kind of thing.  All we use JSCS for is to make indentation consistent, because that really affects readability.  IMO, stuff like this doesn't really inhibit readability or maintainability.

rheh:
Yeah, I can see your point - a few years back I felt the same.  I feel having a coding standard and publishing one sets good intentions and can encourage those of a caution nature to contribute - which from my own experience can be a little daunting.

dougmoscrop:
If you're going to adopt a style, having an .editorconfig would be helpful. (JSCS also does an 'auto-fix' now, right?)

megawac:
I was going to do this originally in https://github.com/caolan/async/pull/837 along with consistent brace&parenthesis use, but dropped it after resistance. +0 from me

megawac:
If we are going to accept this I think you should set the jscsrc config to use a stricter set of presets (airbnb or google for instance)

aearly:
If we do this, I think we should set up ghooks to run any linters/style-checkers on commit.  Nothing is more annoying when a PR fails CI due to style nits.

hstonge:
+1 for .editorconfig file

PR #907, opened 2015-09-11
apply thisArg

moander:


aearly:
If you need a `thisArg` when using `apply`, just use `bind()`.  `async.apply(myFunc.bind(this), arg1, arg2)` is the same as `myFunc.bind(this, arg1, arg2)`.

PR #909, opened 2015-09-11
next utils

moander:


aearly:
Not really a fan of this.  For functions that pass a variable number of arguments in a waterfall, I'm more a fan of doing `var callback = _.last(arguments)` and going from there.  Also -- concatenating the extra args with any passed-in args seems a bit strange.  It seems like in most cases you wouldn't care about those spare arguments.

moander:
@aearly Not a fan? Whats not to like? :)

`_.last(arguments)(null, 'foo', 'bar')` 
is exactly the same as 
`async.next(arguments, null, 'foo', 'bar')` 
except for the latter not requiring any extra dependencies.

Not sure what you mean with concatenating the extra args. Are you refering to the nextok method? The concatination there is to add null as the first argument.

`_.last(arguments)(null, 'foo')` 
is the same as 
`async.nextok(arguments, 'foo')`

I agree that the nextok is a bit to lazy, but the next method solves a very common task for people using the async lib.

megawac:
`next` is definitely not the right term for this as `iterators`/`generators` are supported by async, which has a conflicting term. I'm also :-1: on this, seems outside of the scope of async (i.e. not an `async` problem)

PR #928, opened 2015-10-13
add typescript declaration

alex20465:
support to typescript developers, i just added the definition-file from **borisyankov/DefinitelyTyped**


aearly:
Is there a way to test that the type declarations are valid? E.g. we validate that our tests conform to the type declarations, and if we change a method, we can get some sort of warning that we also need to update the type definitions. (Sorry, I'm not really familiar with TypeScript's tooling)

PR #938, opened 2015-10-22
async.queue's task callbacks recursion bug

aaronzhang1990:
async.queue.
once a task finished, the wraped `_next` call `q.process` synchronous, when there has many many tasks, these goes to : **RangeError: Maximum call stack size exceeded**

### DIFF
--- a/lib/async.js
+++ b/lib/async.js
@@ -870,7 +870,7 @@
                 if (q.tasks.length + workers === 0) {
                     q.drain();
                 }
-                q.process();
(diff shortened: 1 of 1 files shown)


aearly:
Sorry, but we're moving in the other direction.  We used to do this, but now we now always assume async task functions in the queue for performance reasons. (#696, and [see note here](https://github.com/caolan/async#synchronous-iteration-functions)) You'll have to use `process.nextTick` in your own queue processing function.

Issue #947, opened 2015-10-29
Make queue behavior easier to customize


adam-26:
Hi, I have a use case with a queue where I check a condition in the queue task, and if the condition is not met I put the task back in the queue (this is a simplification). 

From a performance perspective, it would be much better if the tasks that failed the condition were not processed. However, currently it's very difficult to modify the behavior of the queue without accessing the async source.

I will submit a pull request that allows the queue behavior to be customized without needing to copy/paste snippets from the async source.

Let me know if you have any comments.

aearly:
Would this accomplish what you want?

```js
var q = async.queue(function (task, callback) {
  if (matchesFailureCondition(task)) {
    requeue(task);
    return callback();
  }
  // process as normal...
}, concurrency);

function requeue(task) {
  q.push(task);
}
```

adam-26:
No, I'm trying to avoid having to requeue items, it adds unecessary delay to pending tasks. For an example of what I'm wanting to do, see https://github.com/adam-26/async/commit/6ba827d757416ea6dbde21e501511140c400e7fb

PR #950, opened 2015-10-29
Refactor queue to make it easier to customize behavior

adam-26:
Refer to issue #947.

For an example usage, I have created a 'topical queue', where each queue item is assigned a topic. Each topic can be paused/resumed without impacting other topics. See here: https://github.com/adam-26/async/commit/6ba827d757416ea6dbde21e501511140c400e7fb

### DIFF
--- a/lib/async.js
+++ b/lib/async.js
@@ -852,7 +852,7 @@
         else if(concurrency === 0) {
             throw new Error('Concurrency must not be zero');
         }
-        function _insert(q, data, pos, callback) {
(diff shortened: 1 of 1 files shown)


aearly:
I like that this simplifies the implementation of queue a bit, but I don't really like how it exposes some of the private internals (e.g. `q._getTasksToProcess`).  If we expose them, we'll need tests for them, and should make them public.  Otherwise, if we refactor the `queue` some more (which is likely), we'll inadvertently break people's code.

Issue #957, opened 2015-11-16
Version of async.series where each task receives the results of all previous tasks 


alexpusch:
I frequently encounter a situation where a async.series task need the results of several previous tasks. async.waterfall is insufficient since it only provides the result of the last previous task. async,auto is too verbose, and provides additional, unneeded dependency declaration.

A new function (or an improvment of async.series) would come in handy. Each task would receive a result object containing the results of all previous tasks

Possible names: accumulate, seriesAccumulate

Example: 
```js
async.accumulate({
    one: function(callback){
        setTimeout(function(){
            callback(null, 1);
        }, 200);
    },
    two: function(callback, results){
        // results: {one: 1}
        setTimeout(function(){
            callback(null, 2);
        }, 100);
    },
    three: function(callback, results){
         // results: {one: 1, two: 2}
        setTimeout(function(){
            callback(null, results.one + results.two);
        }, 100);
    }
},
function(err, results) {
    // results is now equal to: {one: 1, two: 2, three: 3}
});
```

aearly:
Have you checked out `async.auto`?  It does something very similar to what you describe.

alexpusch:
As I mentioned, in a simple series scenario async.auto is too verbose, The user have to declare that each step is dependent on the previous one, resulting in those ugly arrays and duplicate step names

```js
async.auto({
    one: function(callback){
        setTimeout(function(){
            callback(null, 1);
        }, 200);
    },
    two: ["one", function(callback, results){
        setTimeout(function(){
            callback(null, 2);
        }, 100);
    }],
    three: ["two", function(callback, results){
        setTimeout(function(){
            callback(null, results.one + results.two);
        }, 100);
    }
   ]
},
function(err, results) {
    // results is now equal to: {one: 1, two: 2, three: 3}
});
``

aearly:
We are going to make some enhancements to auto in the next major release -- always have the callback last, and also add `autoInject` (#608) which infers the dependencies from the parameter names.

I don't like this proposal because it relies on the order of the object keys -- something which is not guaranteed across JS environments (and is only guaranteed in Node for small objects).  With arrays it is possible, but it is too similar to `async.waterfall`.  I'm not sure we need something in between `auto` and `waterfall`, especially if we're adding `autoInject`.

aearly:
A way to accomplish what you want:

```js
function seriesAccumulate(tasks, done) {
  async.reduce(tasks, [], function (results, task, next) {
    task(results, function (err, result) {
[...]

PR #959, opened 2015-11-21
improve async.series to pass previous results to tasks

alexpusch:
In issue #957 I've suggested that each async.series task should receive previous tasks results. I've implemented the change. There is a code duplication issue with the _parallel function, I've [tried to refactor](https://github.com/alexpusch/async/commit/bda91c5564f794c2fd8668d8cf286ef451c1ed87) _parallel to accommodate the change but it hearts readability.

As I described in issue #957, I think this change will make async.series much more useful.

### DIFF
--- a/lib/async.js
+++ b/lib/async.js
@@ -731,7 +731,20 @@
     };
 
     async.series = function(tasks, callback) {
-        _parallel(async.eachOfSeries, tasks, callback);
+        callback = callback || noop;
+        var results = _isArrayLike(tasks) ? [] : {};
+
+        async.eachOfSeries(tasks, function (task, key, callback) {
+            task(_restParam(function (err, args) {
+                if (args.length <= 1) {
+                    args = args[0];
+                }
+                results[key] = args;
+                callback(err);
+            }), results);
+        }, function (err) {
+            callback(err, results);
+        });
     };
 
     async.iterator = function (tasks) {
--- a/test/test-async.js
+++ b/test/test-async.js
@@ -1112,13 +1112,16 @@ exports['series'] = {
                 callback(null, 1);
             }, 25);
         },
-        function(callback){
+        function(callback, results){
+            test.ok(results[0], 1);
             setTimeout(function(){
                 call_order.push(2);
                 callback(null, 2);
             }, 50);
         },
-        function(callback){
+        function(callback, results){
+            test.ok(results[0], 1);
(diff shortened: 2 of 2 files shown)


aearly:
Sorry, but I'm :-1:  on this.  I'd rather not introduce another function where the callback isn't the last parameter for iteration functions (we're going to change auto in the next release too).  It is also a subtle breaking change to `series` too.  If you're pass a function that has an optional first argument in the series, this will break.  I'm talking about a function like this:

```js
function doSomething (options, callback) {
  if (!callback) {
    callback = options;
    options = defaultOptions;
  }
  //...
}
```

A common pattern in Node.  If you are doing this with series, it will break with your change:

```js
async.series([
  function (next) {next(null, "foo"}),
  doSomething,
  //...
], done)
```

`doSomething` will be passed `[callback, ["foo"]]`as args and not work as expected.

Also, it's best to not rely on the key ordering of an object.

megawac:
@aearly In the next major, I would like to change the way we call functions which expect user provided parameters. Instead of `fn(err, ...args, callback)` I think `fn(err, [...args], callback)` would be generally more flexible. Thoughts?

alexpusch:
Since this is indeed a breaking change, how about sending the result object as the first argument to the task? Naturally this could only be added in the next major release.

In addition, async.series already allows to depend on key ordering. Users that want to avoid this and use a tasks array can still enjoy previous tasks results by accessing the results object with the task index.

Not providing access to other tasks results makes async.series much less useful, but as you said async.autoInject will avoid async.auto downsides I've mentioned in #975 and make it a good alternative for my suggestion. (BTW autoInject will probably not work in a production client side using code minimization, right?)

aearly:
@megawac can you explain this more?  (Maybe in a separate issue)

Also, just because we have a major release planned, doesn't mean we have to make  breaking changes to `series`.  I'd prefer to keep `series` dead  simple -- just a way to run independent async functions in series. (most people store any shared data in the closure scope).  It's one of the basic building blocks that hasn't changed since `async`'s inception.  If we want to track the progressive results in a series, I'd rather add a new  `async` method.

As in #957, the implementation wouldn't be that complicated:

```js
function seriesAccumulate(tasks, done) {
  async.reduce(tasks, [], function (results, task, next) {
    task(results, function (err, result) {
      results.push(result);
      next(err, result);
    });
  }, done);
}
```

Issue #962, opened 2015-11-25
.until/.while swallow last result, making .map(.until()) impossible


benfleis:
The result is present, simply not passed.

The change would be 100% backward compatible.

Adding a 1-line diff, w/ 20-line test diff.

aearly:
This was closed through #963 .

PR #963, opened 2015-11-25
pass last result to while/until callback

benfleis:
Addresses https://github.com/caolan/async/issues/962

### DIFF
--- a/README.md
+++ b/README.md
@@ -834,8 +834,9 @@ __Arguments__
 * `fn(callback)` - A function which is called each time `test` passes. The function is
   passed a `callback(err)`, which must be called once it has completed with an
   optional `err` argument.
-* `callback(err)` - A callback which is called after the test fails and repeated
-  execution of `fn` has stopped.
+* `callback(err, [results])` - A callback which is called after the test
+  function has failed and repeated execution of `fn` has stopped. `callback`
+  will be passed an error and any arguments passed to the final `fn`'s callback.
 
 __Example__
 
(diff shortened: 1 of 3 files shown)


lib/async.js
```diff
@@ -781,7 +781,7 @@
                 } else if (test.apply(this, args)) {
                     iterator(next);
                 } else {
-                    callback(null);
+                    callback(null, args[0]);
```
thanodnl: Is there a reason to only pass on 1 argument instead of all the arguments that were passed to the `cb`?

For example `waterfall` passes on all the arguments from `cb` on to the next function.

benfleis:
addressed @thanodnl 's suggestion, and tidied tests a bit.

megawac:
:+1:

aearly:
:+1: a nice enhancement.

benfleis:
@megawac @aearly @thanodnl
Just added to the PR/branch, with an updated README to reflect the changes.

README.md
```diff
@@ -834,8 +834,9 @@ __Arguments__
 * `fn(callback)` - A function which is called each time `test` passes. The function is
   passed a `callback(err)`, which must be called once it has completed with an
   optional `err` argument.
-* `callback(err)` - A callback which is called after the test fails and repeated
-  execution of `fn` has stopped.
+* `callback(err, [results])` - A callback which is called after the test fails
+  and repeated execution of `fn` has stopped. This will be passed the results
+  of the last call to `fn`'s callback.
```
megawac: > A callback which is called after the test fails and repeated execution of `fn` has stopped. This will be passed the results of the last call to `fn`'s callback.

This doesn't read so well, maybe
> A callback which is called after the test function has failed and repeated execution of `fn` has stopped. The `callback` will be passed an error and any arguments passed to the final `fn`'s callback.

benfleis: Thanks, should have read that out loud first :)

Update coming.

PR #964, opened 2015-11-26
parallel optional callback NOT optional

YanivSa:
Removed `optional` words


megawac:
Isn't it optional https://github.com/caolan/async/blob/master/lib/async.js#L709

aearly:
Yeah, it's now optional, but it wasn't in the past.

PR #979, opened 2015-12-15
Updated README.md

marcolino:
Added two waterfall examples: with named functions, and with named functions passing arguments to the first one.


aearly:
LGTM :+1:

PR #983, opened 2015-12-18
whilst callback does not return [results] argument

toanalien:


aearly:
This change was unpublished.  The readme should be correct now.

PR #993, opened 2015-12-30
auto stops after error #988

### LINKED ISSUE #988: auto allows functions to start after error is passed
After a task in an `async.auto` calls back with an error (in the following example, this is task `x`) it should stop the processing of future tasks.  However, it still allows the task `c` to run.

Example:

```javascript
async.auto({
  a: function(next) {
    console.time('a:Time');
    console.log('a()');
    setTimeout(function() {
      next(null, 'one');
      console.timeEnd('a:Time');
    }, 200);
  },
  b: function(next) {
    console.time('b:Time');
    console.log('b()');
    setTimeout(function() {
      next(null, 'two');
      console.timeEnd('b:Time');
    }, 50);
  },
  c: ['a', 'b', function(next) {
    console.time('c:Time');
    console.log('c()');
    next(null, { three: 3 });
    console.timeEnd('c:Time');
  }],
  x: function(next) {
    console.time('x:Time');
    console.log('x()');
    setTimeout(function() {
      next(new Error('x'));
      console.timeEnd('x:Time');
    }, 100);
  }
}, function(err, results) {
  console.log('callback(%s, %j)', err, results);
});
```

Output:

```
a()
b()
x()
b:Time: 56ms
callback(Error: x, {"b":"two"})
x:Time: 101ms
a:Time: 205ms
c()
c:Time: 0ms
```

A quick explanation of what is happening (and what should be happening): `a`, `b`, and `x` immediately begin executing (since they have no dependencies).  `b` completes first and `x` calls-back with an error.  At this point the final callback properly executes followed by `x` and `a`. The following execution of `c` is the failure because the error from `x` is called-back before `a` (the dependency of `c`) can complete.

That said, I can see a valid argument for this functionality: i.e. because `c` does not depend on `x` it should be allowed to execute; however, because the callback runs immediately upon receiving an error, can only handle a single error, and does not include the results from `c` it seems that the `auto` function should dump any remaining tasks.

gr2m:
- [x] add test
- [x] fix bug

closes #988

### DIFF
--- a/lib/async.js
+++ b/lib/async.js
@@ -528,6 +528,8 @@
         var results = {};
         var runningTasks = 0;
 
+        var hasError = false;
+
         var listeners = [];
         function addListener(fn) {
             listeners.unshift(fn);
@@ -550,6 +552,7 @@
         });
 
         _arrayEach(keys, function (k) {
+            if (hasError) return;
             var task = _isArray(tasks[k]) ? tasks[k]: [tasks[k]];
             var taskCallback = _restParam(function(err, args) {
                 runningTasks--;
@@ -562,6 +565,8 @@
                         safeResults[rkey] = val;
                     });
                     safeResults[k] = args;
+                    hasError = true;
+
                     callback(err, safeResults);
                 }
                 else {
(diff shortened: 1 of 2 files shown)


aearly:
This looks good.  Nice catch.  :+1:

gr2m:
Glad to help :) Could you release a patch version?

PR #994, opened 2016-01-01
Add some information to error message

mchapman:
I had to put this change in to spot where my typo was, so thought I would make a PR


aearly:
:+1:

PR #996, opened 2016-01-03
Modularization

megawac:
A work in progress to bring modular support and lodash implementations of certain methods. 

Previous discussion  be found here #984.

Resolves #984

See https://github.com/caolan/async/tree/modularization/build for the generated bundles

-------- 

TODOS

- [x] script releases:
    - generate modular builds/files
    - copy ES modules to build/es
    - add git commit
    - `npm test`
    - smoke test build files
    - `cd build/`
    - publish npm release
    - revert modular build changes, returning async to normal folder structure
    - git tag, git push
- [x] ~~possibly? create npm release for each module~~     
- [x] https://github.com/caolan/async/pull/984#issuecomment-168450571
- [x] use `baseOrderBy` in the implementation of `internal/flatten`?
- [x] ~~move documentation to jsdoc format (bonus?)~~ *save for later*
- [x] test package with npm pack before first 2.0 publish


megawac:
I have managed to remove most of the string parsing logic included by lodash -- the only string parsing code remaining is `toNumber` included by `baseRange`. @jdalton may consider moving the `toNumber` logic to `range` and having `baseRange` assume it is provided numeric `start`, `end` and `step`?

Also @jdalton it'd be nice if there was a stable ref I could link for a `lodash-es` submodule -- it's pretty frustrating to update the submodule ref every couple days

jdalton:
> may consider moving the `toNumber` logic to range and having `baseRange` assume it is provided numeric `start`, `end`, and `step`?

Good idea!

> it'd be nice if there was a stable ref I could link for a `lodash-es` submodule -- it's pretty frustrating to update the submodule ref every couple days

Lodash v4 will be released on the 12th.

megawac:
Cool, so there will be a stable es-4.0.0 ref in v4?

jdalton:
> Cool, so there will be a stable es-4.0.0 ref in v4?

Yep, just as there's a stable 3.10.1 ref.

aearly:
What does the testing situation look like?  Do we just test the bundle?  It might be handy to test the es6 modules directly somehow as well, just to verify that Async works in both forms.

megawac:
@aearly nope, we test [lib/index](https://github.com/caolan/async/blob/modularization/lib/index.js) see https://github.com/caolan/async/blob/modularization/test/test-async.js#L7

jdalton:
Updated the `es` branch with `toNumber` removed from `baseRange`. The gzipped minified size is now 6.17 kB so a 1.9 kB difference which I think is fine.

megawac:
Nice! :+1: that is totally reasonable in my opinion, thoughts @aearly

jdalton:
I actually found another cheap win which will bring the size down to 5.73 kB the diff to 1.4 kB :P

aearly:
Really excellent :+1: .  Lodash is totally worth 1.4kb :stuck_out_tongue: .

The only danger I see is relying on an internal lodash function, but those are pretty stable, right?

jdalton:
> The only danger I see is relying on an internal lodash function, but those are pretty stable, right?

It depends. When relying on internals I usually avoid the `^` range and go with the `~`. I can review the deps and report back if I see anything of concern.

jdalton:
Off hand I think the internals used are stable.

jdalton:
Just checked and the base methods used are totally stable. You should have no problems with them.
I also simplified the deps of `toNumber`.

aearly:
I'm thinking of a clean way to publish the individual method files without polluting the base directory.

My initial thought is to copy the package.json/readme/changelog/etc. to the `build/` directory and run npm publish from there.  We wouldn't have to use the `files` property of the package.json because we would just publish everything in that folder.  This has some drawbacks:

1. We can't use `xyz` any more since it relies on `npm test`, which needs to be run from the root directory.  We'd have to use a custom script to run the tests, bump the versions, and publish.
2. The `package.json` will be out of sync with its location.  The scripts will be assuming it's still in the root, but the actual publish artifacts will be assumed to be in `build/`.  For example, `main` would be just `async-cjs.js` rather than `build/async-cjs.js`, which could be confusing to people submitting PRs.
3. We need to make sure that bower still pulls the correct file.  The bower.json will need to have its `main` remain as `dist/async.js`, and we'll still have to commit that build artifact. The package.json and bower.json `main` fields will now be different.
4. I'd like to publish the ES modules as well as the CJS modules.  I assume we'll copy lib to build and include `"esnext:main": "es/index.js"` or something similar.

Anyone have better ideas?

jdalton:
Anything I can help with. Lodash v4 is out and has stabilized.

aearly:
Nope, not really now that v4 is on npm and we've migrated away from deprecated methods.

Any opinion on the proposed publish process in my last comment?

jokeyrhyme:
Now that https://github.com/rollup/rollup exists, is it worth putting much effort into this at a package level? It would seem that migrating to ES2015 modules here and using an ES2015-aware toolchain is a much easier way to offer consuming projects this level of optimisation.

jdalton:
@aearly 
> Any opinion on the proposed publish process in my last comment?

Why not just copy files to a staging directory during the prepublish. Not sure anything needs to be moved permanently from root (package.json, etc.).

aearly:
@jokeyrhyme We're trying to support both a monolithic CJS or UMD-style import, as well as the individual ES modules, in the same package.  We're authoring them as ES modules, and then using babel and rollup to compile them to other formats.  The goal is to support ES2015 users, as well as existing projects using CJS in node natively.

aearly:
Nearing completion!  I think i have the build and publish process where I want them.  I'd love to have your feedback.

I think the next step is to start cherry picking in changes from master.

TrySound:
I think you should add built files to .gitignore. This will reduce PR a bit ;)

megawac:
Agreed, but lets do it as a final step so people can sanity check the output

aearly:
I think we'll need to keep `build/async.js` checked in in at least, for Bower support.  Or we could just go full lodash and drop Bower support... :new_moon_with_face:

jdalton:
> Or we could just go full lodash and drop Bower support... :new_moon_with_face:

See https://github.com/lodash/lodash/issues/1733#issuecomment-175072673.

aearly:
Ah, I see you dropped explicit support for it, but it still works due to how lodash is authored.  That wouldn't work here, unless we checked in `build/async.js`.

megawac:
I think we should just check in the `/dist` folder with the built async.js
and async.min.js files

aearly:
I...think we're ready to merge.  I just merged from master, and resolved the conflicts.  Luckily, they were pretty minor.   I think after this we can finish the rest of the work for the 2.0 milestone on master.

megawac:
I'll review tomorrow @aearly

Makefile
```diff
@@ -1,37 +1,99 @@
...
+	cp $< $@
+
+$(DIST)/async.min.js: $(UMD_BUNDLE_MIN)
+	cp $< $@
 
+build-es: $(ES_MODULES)
+
+$(BUILDDIR)/es/%.js: lib/%.js
+	mkdir -p "$(@D)"
+	sed -r "s/(import.+)lodash/\1lodash-es/g" $< > $@
```
megawac: Nice. Really liking this makefile @aearly

build/async.js
```diff
@@ -0,0 +1,1225 @@
+'use strict';
```
megawac: Do we want this file to be called `build/async` or `build/index`. Also on publish are we going to move the `/build` files to `/` so people can do `import each from 'async/each`?

megawac: I'd prefer `index.js` thoughts @aearly

build/async.min.js
```diff
@@ -0,0 +1,2 @@
...])))});return o.memo=r,o.unmemoized=n,o},P.unmemoize=function(n){return function(){return(n.unmemoized||n).apply(null,arguments)}},P.times=A(P.map),P.timesSeries=A(P.mapSeries),P.timesLimit=function(n,t,e,r){return P.mapLimit(f(n),t,e,r)},P.seq=function(){var t=arguments;return m(function(e){var r=this,u=e[e.length-1];"function"==typeof u?e.pop():u=n,P.reduce(t,e,function(n,t,e){t.apply(r,n.concat([m(function(n,t){e(n,t)})]))},function(n,t){u.apply(r,[n].concat(t))})})},P.compose=function(){return P.seq.apply(null,Array.prototype.reverse.call(arguments))},P.applyEach=T(P.eachOf),P.applyEachSeries=T(P.eachOfSeries),P.forever=function(t,e){function r(n){return n?i(n):void o(r)}var i=u(e||n),o=z(t);r()},P.ensureAsync=z,P.constant=m(function(n){var t=[null].concat(n);return function(n){return n.apply(this,t)}}),P.wrapSync=P.asyncify=function(n){return m(function(t){var e,r=t.pop();try{e=n.apply(this,t)}catch(u){return r(u)}U(e)&&"function"==typeof e.then?e.then(function(n){r(null,n)})["catch"](function(n){r(n.message?n:new Error(n))}):r(null,e)})},"object"==typeof module&&module.exports?module.exports=P:"function"==typeof define&&define.amd?define([],function(){return P}):C.async=P}();
```
megawac: This file should be deleted and moved to `/lib`

build/async.js
megawac: Ah nevermind, I get it -- we publish the `build/` folder.

My comments remain - I'd rather have a `dist/` folder under `build/` with the built `async.js` and `async.min.js` files.

I'm not sure if I want `es/` to be published to npm. I think I would prefer it to be published in a second repo (`async-es` or similar)

jokeyrhyme:
Travis CI tests are failing for Node.js 0.10.x:

>FAILURES: Undone tests (or their setups/teardowns): 
>- memoize - avoid __proto__ key return undefined

Is it time to discuss ending support for Node.js 0.10.x?

support/xyz.sh
```diff
@@ -0,0 +1,171 @@
+#!/usr/bin/env bash
```
megawac: +

support/build/plugin-lodash-import-rename.js
```diff
@@ -0,0 +1,24 @@
+import _ from 'lodash';
```
megawac: do we still need this?

megawac:
> Is it time to discuss ending support for Node.js 0.10.x?

I don''t think so, lets work around it.

@aearly left some comments

Makefile
aearly: :smile: I'm totally a Make fanboy.

This does force windows maintainers to use something like MinGW, though.

build/async.js
aearly: This is the UMD build.  I called it `async.js` because it is the most analagous to the `lib/async.js` in v1.x and prior.  The CJS build is called `index.js`.

I'm leaving the `dist/` folder in the root so it can be easily targeted by bower users.  Moving it to `build/dist/` is no problem.

The `es/` folder is an experiment.  I'm copying what lodash 4.x does with `lodash/fp`.  You can `require('async/waterfall')` or `import waterfall from 'async/es/waterfall'` with the same package.

build/async.min.js
aearly: `lib/` or `dist/`?  I do wonder if there is value in having the minified build on npm.

support/build/plugin-lodash-import-rename.js
aearly: Yes, it changes our `import foo from 'lodash'` to use the correct ES version when performing the rollup.

aearly: Maybe I should write a comment.

jdalton:
@jokeyrhyme 
> Is it time to discuss ending support for Node.js 0.10.x?

See stats from npm
https://twitter.com/seldo/status/672569831565078528
and Semaphoreci
https://twitter.com/JavaScriptDaily/status/677741377745240064

jokeyrhyme:
@jdalton woah. Depressing, but good to know!

aearly:
Yep, lots of organizations haven't had time to migrate.  And don't be sad, be glad they're not on 0.8!

aearly:
Remaining questions:

- Should we publish ES modules as `'async/es/foo'` or in a separate package `'async-es/foo'`?
- Where should `async.js` and `async.min.js`(UMD builds) go?
- `build/dist/` or `dist/` (for bower UMD support)?

build/async.js
megawac: Yeah, I think I'd prefer to keep them in seperate repos for build size concerns.

I was suggesting `/dist/async.js` for organizational reasons (for instance if we decided to add an `async.async` function in the future it wouldn't be possible due to the organization of built files.

build/async.min.js
megawac: Whoops `dist/`* and there might be if people do

```js
<script src="./node_modules/dist/async.min.js"></script>
```

megawac:
@jdalton was 0.8 not included in that list?

megawac:
> Should we publish ES modules as 'async/es/foo' or in a separate package 'async-es/foo'?

I'd say separate modules for build size concerns -- few people will likely be using our `/es` stuff

> Where should async.js and async.min.js(UMD builds) go?  build/dist/ or dist/ (for bower UMD support)?

I would say both. Add a `/dist` folder so people can download directly off github and then publish a `/build/dist` folder as well so people can use the compiled file on node.

Question is should we make the mainfile for the repo `dist/async.js` or `index.js`?

Also we only publish the build folder to npm correct @aearly?

Makefile
megawac: Thats fine, as long as they can write and run tests its okay for me

charlierudolph:
`lib/async.js` needs to be removed.

jdalton:
@megawac
> was 0.8 not included in that list?

Not in semaphore's in the npm report 0.8 had 0.1%.

aearly:
@megawac correct, we only publish what is in build.  I'm going to separate out the ES modules into a separate folder so it can be a separate package.

aearly:
https://www.npmjs.com/package/async-es

May as well claim it now. :P

aearly:
Okay, I cleaned up the UMD `dist/` and separated out the ES build.  *finger moves closer to merge button*

megawac:
I'll review again tonight

megawac:
Havent checked yet but do we test the compiled builds Alex?

aearly:
We test `lib/` with nodeunit and mocha using babel to compile the ES modules.  We have a simple smoke-test for the compiled output (CJS, UMD, and ES formats).

support/module_template.md
```diff
@@ -0,0 +1,13 @@
+# async.<%= name %>
```
megawac: Do we want to keep this?

aearly: I don't think so.  It's only there if we were to publish each function as its own package. (same thing with the Gulpfile)

aearly:
Anymore comments?  I suppose anything else could be fixed on master.

jtwebman:
Lets merge it, it is big change :)

megawac:
Do we want to add the build directories to gitignore first?

aearly:
I guess now that the code review is complete we can ignore them.

megawac:
The last question in my mind is what do we want our main file to be? The
built file (to avoid imports) or the CJS file (current)

aearly:
The CJS index seems more "correct", but it would only be better in node-land.  If you're browserify-ing the UMD build will be smaller, and it will still wok in node.  I say we make dist/async.js the main.

```
$ browserify --standalone async build/index.js | uglifyjs -mc | gzip -9 | wc -c
11089

$ cat dist/async.min.js |  gzip -9 | wc -c
6680
```

megawac:
For the browserify case, it also depends on whether the user is importing
modules which also depend on lodash.

aearly:
If the browserify user was cherry-picking async methods, they'd get some overlap with lodash and some possible de-deuping, if they were also cherry-picking lodash methods.  But if you require both async and lodash wholesale, you'll get no module overlap, and some code duplication.  It's a tradeoff between code duplication and CJS module overhead.

jdalton:
Just popping in, if you use browserify, are you using bundler-collapser? I know webpack generally produces smaller builds and has more flexibility in terms of module path hackery.

aearly:
```
$ browserify --standalone async build/index.js | bundle-collapser | uglifyjs -mc | gzip -9 | wc -c
9393
```

Helps a little.

jtwebman:
I thought browserify was going away. It seems more and more people only are using npm. Maybe that is just from what I have seen. Has anyone else noticed it?

aearly:
Browserify is still alive and well, but whatever the case, here it's just a quick way to get a feel for what the module overhead is, regardless of what bundler you're using.

megawac:
When do you want to release v2? and do we want to do a pre-release?

martinheidegger:
I am jumping on this bandwaggon waaay too late but: Why all this effort of splitting it out to own packages? This will make npm just slower. npm supports `require('async/something') for files within the async repo. Async only needed to distribute the modules as `.js` files in the tree:

````
- package.json
- parallel.js
- each.js
- serial.js
- ....
```
Then you could do `require('async/parallel')` would automatically just load the code required for parallel. This would significantly faster in every respect, wouldn't it?

megawac:
>  Why all this effort of splitting it out to own packages? 

We're not doing that at the moment, but the benefit is the download for an individual method is smaller than the entire async package

jdalton:
Though with npm cache packages of a given version are only downloaded once. Subsequent requests hit the local npm cache on your system.

aearly:
I want to release 2.0 when all the issues in the milestone are complete. (All the breaking changes). I'd also like to get the docs migrated to jsdoc and the tests ported to Mocha, but those aren't required.

martinheidegger:
@megawac I do understand the need for splitting out the modules in own files/modules as this will reduce the size if used by packages such as a browserify but when downloading the files through npm this method should slow down things (considering the package.json and other overhead). Browserify (and babel afaik) do support deep binding `require("<package>/<module>")` so the bundled files would already become a lot smaller.

megawac:
> rowserify (and babel afaik) do support deep binding require("<package>/<module>") so the bundled files would already become a lot smaller.

Sure, assuming that none of the user's depencies require `async` and do `require("async")`.

Also, at the moment babel does nothing in regards to `import <package>/<module>` besides transpile it to `require("<package>/<module>")`

@aearly I think it makes sense to set a package.json field for `"browser": "index.js"`, the reason being is webpack and rollup support tree shaking. This would allow the bundler to exclude unused modules (as if the user did `require('async/some-module')`

martinheidegger:
@megawac Users likely will always be able to do `require("async")` and blow up their dependencies.

megawac:
For reference, I measured the require time of using `index.js` vs `dist/async` in v2.

|Env      | dist/async.js   | async/index.js  |
|---      |---              |---              |
|v0.12    | 8 ms            | 69ms            |
|v4.4.7   | 9 ms            | 58ms            |
|v6.2.2   | 11 ms           | 46ms            |

<details>
  <summary><h4>Methodology of test</h4></summary>


  I used [time-require](https://www.npmjs.com/package/time-require) to make these measurements. I ran a tiny script

  ```js
  require('time-require');

  var async = require('async');
  var async1 = require('async/index');
  ```

  I ran this in v0.12, v4, and v6 installed via nvm 5 times each and took the average (tended not to deviate much) on my laptop

  <h6 class="title">Computer</h1><table><tr><td colspan="2" class="stitle">Summary</td></tr>
  <tr><td colspan="2" class="sstitle">Computer</td></tr>
  <tr><td class="field">Processor</td><td class="value">4x Intel(R) Core(TM) i5-4300M CPU @ 2.60GHz</td></tr>
  <tr><td class="field">Memory</td><td class="value">11988MB (7240MB used)</td></tr>
  <tr><td class="field">Operating System</td><td class="value">Ubuntu 14.04.4 LTS</td></tr>
</details>

<details>
  <summary>Costs involved loading async/index (node 6)</summary>

Note: I snipped a couple irrelevant parts of this log
  ```
Start time: (2016-07-13 22:37:29 UTC) [treshold=1%]
 #  module                          time  %
 5  ./isFunction (...sFunction.js)   1ms  ▇ 2%
 6  ./toNumber (no.../toNumber.js)   1ms  ▇ 2%
 7  ./toFinite (no.../toFinite.js)   1ms  ▇ 2%
 8  ./toInteger (n...toInteger.js)   1ms  ▇ 2%
 9  lodash/rest (n...dash/rest.js)   2ms  ▇▇ 4%
10  ./internal/app...applyEach.js)   3ms  ▇▇ 6%
11  ./_getLength (...getLength.js)   1ms  ▇ 2%
12  lodash/isArray...ArrayLike.js)   1ms  ▇ 2%
13  ./getIterator...tIterator.js)    1ms  ▇ 2%
14  ./isArray (nod...h/isArray.js)   1ms  ▇ 2%
15  ./_indexKeys (...indexKeys.js)   2ms  ▇▇ 4%
16  lodash/keys (n...dash/keys.js)   2ms  ▇▇ 4%
17  ./iterator (no.../iterator.js)   4ms  ▇▇▇ 8%
18  ./eachOfLimit...chOfLimit.js)    5ms  ▇▇▇ 10%
19  ./internal/doP...llelLimit.js)   6ms  ▇▇▇▇ 12%
20  ./mapLimit (no.../mapLimit.js)   6ms  ▇▇▇▇ 12%
21  ./map (node_mo...async/map.js)   6ms  ▇▇▇▇ 12%
22  ./applyEach (n...applyEach.js)   9ms  ▇▇▇▇▇▇ 18%
23  ./applyEachSer...achSeries.js)   1ms  ▇ 2%
24  ./asyncify (no.../asyncify.js)   1ms  ▇ 2%
25  ./_baseFor (no.../_baseFor.js)   1ms  ▇ 2%
26  lodash/_baseFo...aseForOwn.js)   1ms  ▇ 2%
27  ./auto (node_m...sync/auto.js)   1ms  ▇ 2%
28  lodash/_arrayM..._arrayMap.js)   1ms  ▇ 2%
29  ./_checkGlobal...eckGlobal.js)   1ms  ▇ 2%
30  ./_root (node_...ash/_root.js)   1ms  ▇ 2%
31  ./_Symbol (nod...h/_Symbol.js)   1ms  ▇ 2%
32  ./_baseToStrin...eToString.js)   1ms  ▇ 2%
33  ./_stringToArr...ngToArray.js)   1ms  ▇ 2%
34  lodash/trim (n...dash/trim.js)   2ms  ▇▇ 4%
35  ./autoInject (...utoInject.js)   4ms  ▇▇▇ 8%
36  ./internal/que...nal/queue.js)   1ms  ▇ 2%
37  ./cargo (node_...ync/cargo.js)   2ms  ▇▇ 4%
38  ./eachOfLimit...chOfLimit.js)    1ms  ▇ 2%
39  ./eachOfSeries...hOfSeries.js)   1ms  ▇ 2%
40  ./reduce (node...nc/reduce.js)   1ms  ▇ 2%
41  ./seq (node_mo...async/seq.js)   1ms  ▇ 2%
42  ./compose (nod...c/compose.js)   1ms  ▇ 2%
43  ./internal/doP...oParallel.js)   1ms  ▇ 2%
44  ./concat (node...nc/concat.js)   1ms  ▇ 2%
45  ./internal/doS.../doSeries.js)   1ms  ▇ 2%
46  ./concatSeries...catSeries.js)   1ms  ▇ 2%
47  ./constant (no.../constant.js)   1ms  ▇ 2%
48  lodash/identit.../identity.js)   1ms  ▇ 2%
49  ./internal/cre...ateTester.js)   1ms  ▇ 2%
50  ./detect (node...nc/detect.js)   2ms  ▇▇ 4%
51  ./doDuring (no.../doDuring.js)   1ms  ▇ 2%
52  ./eachLimit (n...eachLimit.js)   1ms  ▇ 2%
53  ./each (node_m...sync/each.js)   1ms  ▇ 2%
54  ./ensureAsync...sureAsync.js)    1ms  ▇ 2%
55  ./filter (node...nc/filter.js)   1ms  ▇ 2%
56  ./forever (nod...c/forever.js)   1ms  ▇ 2%
57  ./mapValues (n...mapValues.js)   1ms  ▇ 2%
58  ./memoize (nod...c/memoize.js)   1ms  ▇ 2%
59  ./internal/par.../parallel.js)   1ms  ▇ 2%
60  ./parallelLimi...llelLimit.js)   1ms  ▇ 2%
61  ./parallel (no.../parallel.js)   1ms  ▇ 2%
62  ./queue (node_...ync/queue.js)   1ms  ▇ 2%
63  ./priorityQueu...rityQueue.js)   1ms  ▇ 2%
64  ./reduceRight...duceRight.js)    2ms  ▇▇ 4%
65  ./reject (node...nc/reject.js)   1ms  ▇ 2%
66  ./rejectSeries...ectSeries.js)   1ms  ▇ 2%
67  ./retryable (n...retryable.js)   1ms  ▇ 2%
68  ./some (node_m...sync/some.js)   1ms  ▇ 2%
69  ./timeout (nod...c/timeout.js)   1ms  ▇ 2%
70  ./timesSeries...mesSeries.js)    1ms  ▇ 2%
71  ./internal/onl...nal/onlyOnce)   1ms  ▇ 2%
72  ./whilst (node...nc/whilst.js)   1ms  ▇ 2%
73  ./until (node_...ync/until.js)   1ms  ▇ 2%
74  async/index (n...ync/index.js)  44ms  ▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇ 86%
Total require(): 334
Total time: 51ms
  ```
</details>

aearly:
Cool!  Thanks for backing up my gut assertions with some actual data. :smile:

.babelrc
```diff
@@ -0,0 +1,3 @@
+{
+  "presets": ["es2015"]
```
melvinmezzy, line 2: my information

melvinmezzy (commented):
i need my information to run as my new device im on

support/build/plugin-lodash-import-rename.js
melvinmezzy: yes for my new device im on

PR #1002, opened 2016-01-07
Use `nonexistent` instead of `inexistant` in error message.

mattcollier:
Inexistant is not a commonly used English word.


aearly:
LGTM

Issue #916, opened 2015-09-23
npm test fails on windows - jshinit does not recognize lib/*.js


smartmouse:
On a command prompt windows , with node 0.12.7
When I invoke npm test,   jshinit cannot find lib/*.js

However, if I modify  lib/*.js to lib/async.js without the wildcard, it works.
Are you aware of this behaviour on windows.
Is this jshint issue on windows?
On linux , this is working fine.

 "scripts": {
    "mocha-node-test": "mocha mocha_test/",
    "mocha-browser-test": "karma start",
    "mocha-test": "npm run mocha-node-test && npm run mocha-browser-test",
    "nodeunit-test": "nodeunit test/test-async.js",
    "test": "npm run-script lint && npm run nodeunit-test && npm run mocha-test",
    "lint": "jshint lib/async.js test/test-async.js perf/benchmark.js && jscs lib/async.js test/test-async.js perf/benchmark.js",
    "coverage": "nyc npm test && nyc report",
    "coveralls": "nyc npm test && nyc report --reporter=text-lcov | coveralls"
  },

megawac:
Feel free to send a PR changing this. This is a long time issue on windows

aearly:
Closed by #1003

PR #1003, opened 2016-01-09
Fix Lint for Windows by removing unnecessary wildcarding

proflayton:
https://github.com/caolan/async/issues/916
The wildcards are not necessary and causes sadness for Windows users.


aearly:
:+1:

PR #1028, opened 2016-02-16
filter, reject, detect, some, every with error

charlierudolph:
@megawac 

#774 but against modularization branch


megawac:
lgtm, /cc @aearly

aearly:
:+1:

PR #1029, opened 2016-02-19
It just feel weird to put q.push here

SCLeoX:
It is a method of object `queue` and `q` is just a instance of object `queue` in the example.

I know... It's just a very tiny change but it makes it a tiny little better. XD

Also, `queue.push()` seems not mentioned in method list.


aearly:
Eh, it's still a bit strange.  In the examples, we use `queue` to refer to the function, and `q` to refer to an instance returned from `queue`.

PR #1034, opened 2016-02-24
add unsaturation event to `priorityQueue`

suguru03:
Related to https://github.com/caolan/async/pull/1030

I fixed `priorityQueue`.
It has `unsaturated` params but it doesn't work.
Could you check this PR?


aearly:
Nice.  Good catch.

suguru03:
thanks!

PR #1040, opened 2016-02-27
Set "browser" to index to allow tree-shaking

megawac:
Brought this up in the main PR but probably better suited for another issue (here ^^). The way we've restructured the repo its well suited for tree-shaking features in webpack and rollup (and other bundlers in the future). This would ease the bundlers ability to cherry-pick only modules used (for instance `forEachOf` as our `index` is just `import` -> `export`. It would also enable lodash modules to be cherry-picked once.

Thoughts @aearly ?

Also should the `/index.js` file point to `lib/index` rather than `build/index` as `build` may or may not exist

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "description": "Higher-order functions and common patterns for asynchronous code",
   "version": "2.0.0-alpha",
   "main": "dist/async.js",
+  "browser": "index.js",
   "author": "Caolan McMahon",
   "repository": {
     "type": "git",


aearly:
We already set main to `index.js` in `support/sync-es-package.js`.  `dist/async.js` isn't included at all in the ES module (or at least is shouldn't be).

I also think that the `/index.js` should point at `dist/async.js`. It's there so you can have something useful when using the package through git (like we do for the benchmarks).  However, if the package json points at `dist/async.js` already (which we check in), `/index.js` might not be needed at all.

megawac:
Yeah, I'm fine with `index` pointing to the dist file or having `main` simply point to the dist file. Anyway, thoughts on adding a `browser` field point to index?

aearly:
I think you want `dist/async.js` when bundling the whole library (CJS) for the browser.  My experiments in the other thread showed that it's 6kb (vs 9kb for `index.js`) after bowserify/uglify/gzip.  For ES, (where you can do tree-shaking), the main is already `index.js` -- we don't publish `dist/` there.  To take advantage of the modularization for CJS, you have to `require('async/whatever')`.

aearly:
Can Webpack do tree-shaking for CJS?  Thats the only case where you'd want `index.js`.

megawac:
Ah true, tree-shaking isn't supported for CJS I don't think. Is the -es module main set to `index`?

aearly:
Yes.

PR #1042, opened 2016-02-28
Auto argument ordering

### LINKED ISSUE #1023: async.auto: should subsequent tasks still run if the chain as a whole has been marked as failed?
I was under the impression that the behavior of `async.auto` was that if one of the tasks returned an `err`, the global callback would be called with this error and **all** subsequent tasks would not execute. After all, why would they? The global callback has been called so the error has been reported already.

It turns out that only tasks *that depend on the faulty task* will not run, and the rest will.

I'm not sure if it makes sense for `async.auto` to report failure by means of a single global callback on one hand, and still run subsequent tasks on the other hand: as far as the client is concerned, the whole thing failed.

### LINKED ISSUE #1036: Fix callback argument order in async.auto
We should standardize the signature of async functions used in `async.auto`.  Rather than always having the callback first, and the optional `results` object second, we should always make the callback the last argument.

We could also include a `flip` function that swaps the 2 arguments, to ease migration.  (Or just instruct people to `_.rearg(asyncFn, 1, 0)`).

aearly:
Closes #1036 and resolves #1023 

Makes `auto` tasks work like standard async functions -- the callback is always the last arg.


megawac:
Lets get the `some`/`every` error callback stuff in and then do a prerelease @aearly

aearly:
I'd wait until we have all the scheduled breaking changes in.  Then early adopters only have to audit their code once for 2.0.

PanMan:
It really confused me that this is on the main readme, while my code (on 1.x) doesn't handle this.

PR #1049, opened 2016-03-07
Refactor auto to not need a deferral

aearly:
Related to #696 

I had to completely rework things.  Auto should be a bit more efficient now for complicated workflows as well.

This does change execution order slightly, but it's still correct behavior.


aearly:
@megawac Would love a code review here.

Issue #1016, opened 2016-01-28
constant() should support dynamic args


ross-nordstrom:
In `v1.5.2` constant seems to be hard-coded to be invoked with 1 arg: a callback. **This is limiting**

```js
    async.constant = _restParam(function(values) {
        var args = [null].concat(values);
        return function (callback) {
            return callback.apply(this, args);
        };
    });
```

I'd like to propose the inner, curried function take dynamic args as well:

```js
    async.constant = _restParam(function(values) {
        var args = [null].concat(values);
        return _restParam(function (callbackArgs) { // <= We only care about the callback,
            var callback = callbackArgs.pop();      //    and assume it's the last arg
            return callback.apply(this, args);      //    when the curried function is invoked
        });
    });
```

## Why do I care?

Composability.  I don't want to care about the callback-signature of other functions in Async-chains.

```js
async.waterfall([
  doSomething/*(cb)*/, //=> Invokes cb with (err, arg1, arg2)
  async.constant('Pretend doSomething did somethingElse')/*(arg1, arg2, cb)*/
], doneCb);
```

aearly:
I like this -- a simple enhancement.

aearly:
Closed via #1052

PR #1052, opened 2016-03-08
Support dynamic arguments in async.constant (ref. #1016)

ajfranzoia:
Add support for dynamic arguments in constant(), references issue #1016.

### DIFF
--- a/lib/constant.js
+++ b/lib/constant.js
@@ -4,7 +4,8 @@ import rest from 'lodash/rest';
 
 export default rest(function(values) {
     var args = [null].concat(values);
-    return function (cb) {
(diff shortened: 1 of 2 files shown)


test/test-async.js
```diff
@@ -3722,6 +3722,17 @@ exports['constant'] = function (test) {
     });
 };
 
+exports['constant with multiple arguments'] = function (test) {
```
aearly: BTW, you don't need to duplicate the test here.  We're migrating away from nodeunit.

ajfranzoia:
Removed duplicated test.

PR #1055, opened 2016-03-09
autoInject

aearly:
This is adapted from the code in #608.


steverobb:
Glad to see this go in!  I had a new version ready to push, but I've had trouble with running the new build/test frameworks under Windows.

aearly:
Hmmn, the build would require something like MinGW to run on windows, but the tests should work, at least.

PR #1058, opened 2016-03-12
added retryable wrapper for async tasks

aearly:
This splits out the old "control flow" behavior from retry.  It wraps a task, rather than immediate running it.

@megawac would love your code review, as well as perhaps an idea for a better name?


lib/retryable.js
```diff
@@ -0,0 +1,11 @@
+import retry from './retry';
+import rest from 'lodash/rest';
+
+export default function (opts, task) {
+    return rest(function (args) {
+        var callback = args.pop();
+        retry(opts, function (cb) {
+            task(...args.concat([cb]));
```
megawac: remove splat?

aearly: Good catch.

aearly:
What do you think about the `retryable` name?

PR #1060, opened 2016-03-14
update readme - iteratee - callback(truthValue) - has one argument - …

askhogan:
…example has 2


aearly:
See the disclaimer at the top of the readme.

PR #1061, opened 2016-03-14
update readme - iteratee - callback(null, truthValue)

askhogan:
See https://github.com/caolan/async/pull/1060 - either that one is correct or this one is.  Cannot both be true.


aearly:
This is how things will be in 2.0.

PR #1073, opened 2016-03-21
Update README.md

dhritzkiv:
Fix final callback documented arguments for `filter`method, to reflect changes in 2.0. (adds the `err` argument before the `results` argument)


aearly:
Ah, good catch!

PR #1091, opened 2016-03-29
updated info about retry

Yashko:
im using async.retry in my project and dont find information - if you want to continue executing you function, callback with `null` in second argument


aearly:
This isn't quite right -- `retry` will continue until it gets a null `err`, not until the `result` is non-null.

Yashko:
then we need to add this into readme

Yashko:
now i found it, ty

> Attempts to get a successful response from `task` no more than `times` times **before
returning an error**

PR #1094, opened 2016-04-04
Bind Context to Functions wrapped by initialParams

### LINKED ISSUE #1093: ensureAsync overrides "this" binding
Not sure if this is on purpose or not but a "this" is passed into fn.apply inside of ensureAsync which overrides any previous binding of the the function to another context. https://github.com/caolan/async/blob/master/lib/ensureAsync.js#L19.

bspates:
Fixes #1093 
Allow function wrapping methods like ensureAsync, that internally use initialParams, to have a context bound to the function they have wrapped. This allows something like this to work:
```javascript
var async = require('async');

var foo = function(one) {
  console.log(this);
};

foo = async.ensureAsync(foo);
foo = foo.bind({bar: true});

foo(); // prints {bar: true}
```


mocha_test/initialParams.js
```diff
@@ -0,0 +1,26 @@
+import initialParams from '../lib/internal/initialParams';
+var expect = require('chai').expect;
+
+describe('initialParams', function() {
```
aearly: Can you test `ensureAsync` rather than the internal helper function?

mocha_test/initialParams.js
```diff
@@ -0,0 +1,26 @@
+import initialParams from '../lib/internal/initialParams';
+var expect = require('chai').expect;
+
+describe('initialParams', function() {
+  it('should curry any bound context to the wrapped function', function() {
```
aearly: Also, I don't think "curry" is the right term here -- this is just context binding.

aearly:
It also looks like there's a lint error preventing the tests from running.

PR #1012, opened 2016-01-26
Add reflect function wrapper to allow always passing iterations

jtwebman:
Add reflect that wraps a function with a always passing callback and a object with error or value property set.

This is one way to solve issue #942.

PR #1095, opened 2016-04-04
Add back reflect and reflectAll.

jtwebman:
This is the new PR for the reflect for 2.0 (https://github.com/caolan/async/pull/1012). I change the branch because of the big changes to files.


jtwebman:
The tests failed but I don't think it is for my changes?

aearly:
Yep, I re-ran the test and all is good.  Thanks!

Issue #1099, opened 2016-04-05
autoInject doesn't unpack results into the callback


steverobb:
The original version of autoInject unpacked arguments into the final callback.  The current version does not do this:

```
async.autoInject({
    a: function(cb) { cb(null, 1); },
    b: function(cb) { cb(null, 2); }
}, function(err, a, b) {
    console.log('a: ' + a + ' b: ' + b);
});
```

Expected: a: 1, b: 2
Actual: a: [object Object], b: undefined

The documentation still refers to the original behaviour.

steverobb:
To be clear, I see this as an oversight made during the porting of the code, not a documentation error.

aearly:
Closed by #1100

PR #1100, opened 2016-04-06
Fix for autoInject argument injection into final callback

steverobb:
Supports array and non-array forms.

Intended as a fix for issue: https://github.com/caolan/async/issues/1099


aearly:
You're right, this was an oversight.

PR #1103, opened 2016-04-07
Call out louder that this is just an RC, not yet the released version

agibson-vrnt:
Puts a large note right at the top that this isn't the currently released version. I wasted hours the other day wondering why the documentation didn't line up with how .filter works in the current node.js npm version of async.


aearly:
Thanks, but we've already tried making such a callout at the top -- it makes no difference to the number of issues submitted...

agibson-vrnt:
It would have saved me half a day.

megawac:
@aearly we can remove the readme docs soon and use a site which should make this crystal clear :)

PR #1119, opened 2016-04-21
Fixes #1117

### LINKED ISSUE #1117: sourceMappingURL in minified build is not correct
In `async.min.js` there is a reference to the sourcemap:

`//# sourceMappingURL=dist/async.min.map`

However, this reference is not correct: because `async.min.js` is also in the dist-folder. This causes errors when using the script in a browser for example. The sourceMappingURL should be:

`//# sourceMappingURL=async.min.map`

I guess this URL is generated somewhere in a buildscript, but I was not able to find and fix it in a PR.

hargasinski:
Manually setting the source map URL in the `Makefile`,as when `--source-map` is specified without `--source-map-url`, the source map URL defaults to the values passed to `--source-map`. See
[the uglifyJS options](https://github.com/mishoo/UglifyJS2#user-content-usage). 

I didn't want to commit an entire rebuild of `dist` for a one line fix, but if anything, `dist/async.min.js` can be manually edited to change `//# sourceMappingURL=dist/async.min.map` to `//# sourceMappingURL=async.min.map` before the next rebuild.

Fixes #1117

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -62,6 +62,7 @@ $(UMD_BUNDLE_MIN): $(UMD_BUNDLE)
 	mkdir -p "$(@D)"
 	$(UGLIFY) $< --mangle --compress \
 		--source-map $(DIST)/async.min.map \
+		--source-map-url async.min.map \
 		-o $@
 
 $(DIST)/async.js: $(UMD_BUNDLE)


hargasinski:
Sorry about the double pull request, I don't know why there were so many commits in #1118 :(

hargasinski:
I'm looking at the failing test for this PR, it appears that the results for the `queue bulk task` test were returned in the wrong order. It's weird since #1118 is passing even though they are the exact same change.

PR #1125, opened 2016-04-25
Arrow function support in ES6.

steverobb:
Support for ES6 arrow function in autoInject:

```
async.autoInject({
    task1: (cb)        => cb(null, 1),
    task2: (task3, cb) => cb(null, 2),
    task3: cb          => cb(null, 3)
}, (err, task3, task1) => {
    // use task1 and task3
});
```


lib/autoInject.js
```diff
@@ -4,10 +4,10 @@ import arrayMap from 'lodash/_arrayMap';
 import clone from 'lodash/_copyArray';
 import isArray from 'lodash/isArray';
 
-var argsRegex =  /^function\s*[^\(]*\(\s*([^\)]*)\)/m;
+var argsRegex = /^(function[^\(]*)?\(?\s*([^\)=]*)/m;
 
 function parseParams(func) {
-    return  func.toString().match(argsRegex)[1].split(/\s*\,\s*/);
+    return func.toString().match(argsRegex)[2].trim().split(/\s*\,\s*/);
```
megawac: don't use `.trim()` it doesn't have sufficient vendor support. Either import `lodash/trim` or use a different approach. Is it necessary?

mocha_test/autoInject.js
```diff
@@ -74,4 +74,17 @@ describe('autoInject', function () {
         });
     });
 
+    // Needs to be run on ES6 only - not sure how you plan to handle this
```
aearly: I'd be hesitant to add this until we have a good test for it.  It would be nice to do some sort of feature detection and conditionally run this test, but an unsupported arrow function will cause a syntax error before the test even gets evaluated.  Using Babel won't work because it will compile it to a `function () {}` and not really test the feature.   We need a strategy for this.

lib/autoInject.js
jdalton: How far back are you wanting to support?
https://kangax.github.io/compat-table/es5/#test-String_properties_and_methods_String.prototype.trim

jdalton: A good read on coercing functions to strings:
http://perfectionkills.com/state-of-function-decompilation-in-javascript/

steverobb: The trim is necessary because it doesn't support spaces around the function name parameters, e.g. `function( x, y, z )`, and then there was a space after the final parameter.  This wasn't obvious until I added support for arrow functions like `x => y`.  I tried inventing a regex which coped with both functions and arrows, but failed, and the trim seemed like an easy workaround.

Alternatively, it could import a dedicated library, e.g.: https://www.npmjs.com/package/get-parameter-names

jdalton: Can the arg sniffing just be removed altogether?

jdalton: It seems like a good 2.0 move to drop the auto-sniff stuff.

jdalton: or in this case never start (I realize its new in 2).

steverobb: Its entire purpose is to 'auto-sniff' to aid readability, otherwise you'd just use auto.  What is the objection here?

jdalton: It's hacky and brittle and has proved problematic for libs that have done this in the past (Prototype.js and Angular).

steverobb: It's not my library, so it's not my call, but these issues have already been considered:

https://github.com/caolan/async/pull/608#issuecomment-168440002

jdalton: It's a bad move every attempt. The pattern should be avoided.

steverobb: That's one opinion.  But if you wish to see it removed, I think you should raise an issue ticket.  This is not the correct forum.

jdalton: Pssh like srsly whose this jdalton dude. What does he know anyways. Amiright.

steverobb: This is not about you or your reputation.  You have a valid opinion and I encourage you to raise it for discussion.  Just not in the patch comments of this tangentially-related PR.

aearly: Yeah, sniffing parameters is the whole point of `autoInject`.  By using it, you accept all the caveats associated with that approach.  If you don't want it, you can use `auto` (or the array-style form for tasks).

jdalton: 👻

aearly: @jdalton any ideas on how to feature-detect native arrow-function support for testing?

jdalton: Don't forget default params.
```js
function a(b=1) {}
a+''
"function a(b=1) {}"
```

jdalton: Also comments in IE 9 or so:
```js
Refresh the page to see messages that may have occurred before the F12 tools were opened. 
function a(/*b,c,*/d) {} 
a+'' 
// => "function a(/*b,c,*/d) {}" 
```

jdalton: Or bound functions:
```js
function a(b,c) {}
var b = a.bind({})
b+''
// => "function () { [native code] }"

aearly: We'll just document that all of those don't work.

aearly:
One idea I have to feature detect is to use `eval` for the whole test.  If `eval("()=>1")` throws, we know arrow funcs aren't supported.  Then we can do an `eval` on the rest of the test body to test the feature.  Seems really clunky though.

megawac:
Nope, it would syntax error when you load the file. Unless you want to eval
the entire test ^^

aearly:
Yeah, we'd `eval` the whole test....

steverobb:
By the way, though the check-in messages doesn't suggest it, these recent commits utilise eval to allow ES6-only tests.

aearly:
Ah sorry, I didn't notice you had pushed those changes.  It looks like there's now some conflicts -- can you resolve them?

steverobb:
Done!

Issue #1121, opened 2016-04-21
during crashes while passing param to callback fn


WaeCo:
I called the iteratee callback of `during` with an additional parameter and what happened is that this parameter was passed to the `test` function as first parameter.

```
const async=require("async");

var count = 0;
async.during((callback) => {
  console.log(callback);
  callback(null, count < 3);
}, (callback) => {
  count++;
  callback(null, "Some String");
}, (err) => {
  console.log("Done");
});
```

Output:
```
[Function]
Some String
/app/test.js:6
  callback(null, count < 3);
  ^

TypeError: callback is not a function
    at /app/test.js:6:3
    at /app/node_modules/async/lib/async.js:825:22
    at /app/node_modules/async/lib/async.js:167:37
    at /app/test.js:9:3
    at check (/app/node_modules/async/lib/async.js:833:17)
    at /app/test.js:6:3
    at Object.async.during (/app/node_modules/async/lib/async.js:839:9)
    at Object.<anonymous> (/app/test.js:4:7)
    at Module._compile (module.js:398:26)
    at Object.Module._extensions..js (module.js:405:10)
```

Is this intended or a bug?

megawac:
`during` passes any arguments passed by the iteratee function (2nd function) to the test function (1st function). This is intended behaviour.

```js

var count = 0;
async.during(function(callback1) {
  callback1 = arguments[arguments.length-1]
  callback1(null, count < 3);
}, function(callback2) {
  count++;
  callback2(null, "Some String");
}, (err) => {
  console.log("Done");
});
```

WaeCo:
Ok, you should document this

megawac:
If you like you can submit a pull request to clarify the documentation

PR #1129, opened 2016-04-26
Updated description of during #1121

WaeCo:
Updated documentation of `during` in followup of #1121


megawac:
I'm fine with this (=) @aearly?

Issue #1083, opened 2016-03-24
README is too long to npm


hargasinski:
The last four functions (log, dir, noConflict, and timeout) in the README are not shown on npm website.

https://www.npmjs.com/package/async

aearly:
This is tangentially related to #859 .  We've wanted to re-vamp Async's docs for a while -- create a site more similar to Lodash's docs.

hargasinski:
That would be awesome, do you need any help with it? I don't know how far along #859 you are.

aearly:
We haven't finalized the exact building strategy (waiting to see how a similar push for Lodash's docs works out).  At the very least, we want to copy the docs for each method into a JSDoc block in each source file (with full tags and type information), with the intention of parsing them and building a documentation site from them.

megawac:
@hargasinski we could definitely use help with this. If you or anyone else is interested in helping port our docs to [`jsdoc`](http://usejsdoc.org/) compatible formatting in the code it would be greatly appreciated. I recommend doing this method by method. If you're interested I can create a branch for to work off so multiple people can contribute if necessary

Take a look at https://github.com/lodash/lodash/blob/master/lodash.js#L8428-L8470 if you are looking for inspiration.

megawac:
If we get the code in a state where it's documented in JSDoc I can create a docs site similar to ramdas, lazys, or lodashs

hargasinski:
@megawac thank you! I'll look into how lodash does it. I have tomorrow off, so I could probably spend a good amount of time working on it. I'll post a comment on how far I get.

hargasinski:
I may have underestimated the amount of time this is going to take, but I was able to finish the documentation for the collection methods. I should be able to finish the rest tomorrow after work. I created a pull request so you can see what I've done so far, and if anything needs to be changed. I have a couple of quick questions.

1. Since there isn't a real `async` constructor, should I keep including the @memberOf tag?
2. @megawac For the site, how would it be best to document the `iteratee` and `callback` functions? In the readme, the parameters are included the name, such as `iteratee(item, callback)`, but I don't think this will work with jsdoc. I've been doing something similar to the [lodash](https://github.com/lodash/lodash/blob/master/lodash.js#L8453) example you linked, and just including the signature as the last line in the @param description, like `Invoked with (item, callback)`. The jsdoc [site](http://usejsdoc.org/tags-callback.html) recommends to something like:
```js
/**
 * Send a request.
[...]

PR #1135, opened 2016-04-28
add doclets.io config (automated JSDoc pages)

lipp:
Adds documentation to [doclets.io](https://doclets.io) service. Future tags and master branch will be published automatically.

[Preview generated with my account](https://doclets.io/lipp/async/add-to-doclets)

@caolan   If you like this, you are very welcomed to login to https://doclets.io with GitHub OAuth. If you don't like it, I'd be happy to hear what you don't like about it :)

Steps required (3min):
1) @caolan    logs in to doclets.io
2) @caolan   adds async
3) someone merges this PR

Resolves https://github.com/caolan/async/issues/1083

### DIFF
--- a/.doclets.yml
+++ b/.doclets.yml
@@ -1,0 +1,4 @@
+dir: lib
+articles:
+  - Readme: README.md
+  - Changelog: CHANGELOG.md
--- a/lib/doWhilst.js
+++ b/lib/doWhilst.js
@@ -3,7 +3,7 @@
 import whilst from './whilst';
 
 /**
- * The post-check version of [`whilst`](#whilst). To reflect the difference in
+ * The post-check version of {@link async.whilst}. To reflect the difference in
  * the order of operations, the arguments `test` and `fn` are switched.
  *
  * `doWhilst` is to `whilst` as `do while` is to `while` in plain JavaScript.
@@ -11,7 +11,7 @@ import whilst from './whilst';
  * @name doWhilst
  * @static
  * @memberOf async
- * @see `async.whilst`
+ * @see async.whilst
  * @category Control Flow
  * @param {Function} fn - A function which is called each time `test` passes.
  * The function is passed a `callback(err)`, which must be called once it has
--- a/lib/reduce.js
+++ b/lib/reduce.js
@@ -29,15 +29,17 @@ import eachOfSeries from './eachOfSeries';
  * @param {Function} [callback] - A callback which is called after all the
  * `iteratee` functions have finished. Result is the reduced value. Invoked with
  * (err, result).
- * @example
- *
+ * @playground
+ * var async = require('async')
  * async.reduce([1,2,3], 0, function(memo, item, callback) {
  *     // pointless async:
  *     process.nextTick(function() {
  *         callback(null, memo + item)
  *     });
  * }, function(err, result) {
+ *    
  *     // result is now equal to the last value of memo, which is 6
+ *    console.log(err, result)
  * });
  */
 export default function reduce(arr, memo, iteratee, cb) {


aearly:
A cool little service.  @caolan would have to be the one to set it up, since he owns this repo.

We're planning on doing the JSDoc parsing ourselves and setting up the site on http://caolan.github.io/async

caolan:
Looks pretty, but I don't see that it adds much in return for including another third party and spreading our documentation links more thinly. Adding a github pages version of the docs is preferable to this, and is compelling if it's used to include more interactive examples.

-1 on doclets.io

lipp:
@caolan Thanks for your feedback. Actually, interactive examples are supported:
https://doclets.io/lipp/async/add-to-doclets#dl-async-reduce

I can understand your concerns regarding the "spreading" and the "3rd party bloat". 
At the end it "costs" you 3 lines of yml code which eventually save you a lot of time. 

Thanks again for your time and your precious feedback!

megawac:
I actually really dig this assuming it has all the features I was looking for:

- Are [categories or type tags supported](https://github.com/caolan/async/blob/master/lib/auto.js#L32) to define subsections?
- How do you reference other methods, see [do doUntil referencing doWhilst](https://doclets.io/lipp/async/add-to-doclets#dl-async-doUntil)
- We would have to change our @see references
- can the generated jsdoc be exported?

Is it fine if we point a domain to this. I recently purchased http://asyncjs.xyz for $1/y in case we wanted to use it. Otherwise I think we should probably just generate them ourselves

megawac:
Also is is possible to include a javascript file that will be exposed to a clients console. I would want to have `async.js` loaded and the `async` variable exposed in whatever solution we use

lipp:
@megawac The links work if a) backticks are removed b) @link jsdoc is used as documented (see commit).

The @category tag is not part of JSDoc, but as many projects are using it anyways I'll add support for it.

lipp:
As awful as it sounds but may including the doclets page through iframe be an option? The "theme" will be very minimal removing most colors (or make them editable).

This way the doc pages stay "fresh" (webhooks) and get continuously improved as doclets improves.

megawac:
@lipp you mentioned (on pug) it might be possible in the future for the service to publish to gh-pages? If that was the case what would be the real benefit of your service over generating them ourselves?

Also did you see my earlier message?

> Also is is possible to include a javascript file that will be exposed to a clients console. I would want to have async.js loaded and the async variable exposed in whatever solution we use

lipp:
@megawac pushing to the gh-pages could be an option for overcoming the domain/self-hosting issue. I am not sure yet, what is the best solution for "embedding" the doclets content.

No matter which "embedding/push" mechanism will be employed, the point of doclets is, that the docs are **auto-generated** with every push and for all tags/releases. This is much like continuous integration for API docs. Once doclets is enabled, the authors shouldn't care about re-generating docs, link to new versions etc.

As a plus, when doclets gets new features (improved search, navigation, etc) the respective API doc will be "updated" too, since the doc pages are rendered using the "raw jsdoc" data.

Loading a custom script ( script tag + url) would be possible (which would enable libs/modules to be available on console. However, I don't see the advantage over embedding examples through [Tonic](https://tonicdev.com/) or [jsFiddle](https://jsfiddle.net/). Could you elaborate on the advantages using the console please?

Regards and many thanks for the feedback!

megawac:
> Loading a custom script ( script tag + url) would be possible (which would enable libs/modules to be available on console. However, I don't see the advantage over embedding examples through Tonic or jsFiddle. Could you elaborate on the advantages using the console please?

@lipp its nice to be able to load up your console and interact with a library with native autocomplete features and standard debugging tools. For instance, whenever I'm experimenting with lodash/ramda/underscore methods I usually will do it in my console. I usually prefer my console to those interactive tools but having both are nice (like in the [ramda docs](http://ramdajs.com/0.21.0/docs/))

PR #1140, opened 2016-05-03
Fix #1092

### LINKED ISSUE #1092: cyclic dependencies
if we have an object like : 
         obj  =      {
			a : ['c',a],
			b : ['a',b],
			c : ['b',c]
		},
where a,b,c are names of functions, shouldn't this throw cyclic depenedency error if passed to async.auto ?

b-paul:
I believe this does the job requested.

The error text does not report which key(s) are involved in the cycle if one is detected. I haven't spent any thought on that, but will try to figure it out if needed.

### DIFF
--- a/lib/auto.js
+++ b/lib/auto.js
@@ -134,6 +134,11 @@ export default function (tasks, concurrency, callback) {
         uncheckedDependencies[key] = remainingDependencies;
 
         arrayEach(dependencies, function (dependencyName) {
+            if (!tasks[dependencyName]) {
+                throw new Error('async.auto task `' + key +
+                    '` has a non-existent dependency in ' +
+                    dependencies.join(', '));
+            }
             addListener(dependencyName, function () {
                 remainingDependencies--;
                 if (remainingDependencies === 0) {
--- a/lib/auto.js
+++ b/lib/auto.js
@@ -2,7 +2,7 @@
 
 import arrayEach from 'lodash/_arrayEach';
 import forOwn from 'lodash/forOwn';
-import indexOf from 'lodash/indexOf';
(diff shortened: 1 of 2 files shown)


megawac:
Looks good, /cc @aearly

megawac:
Also as we've started using some external dependencies in v2, do you think it makes sense to use an existing dag implementation in auto?

lib/auto.js
```diff
@@ -2,7 +2,7 @@
 
 import arrayEach from 'lodash/_arrayEach';
 import forOwn from 'lodash/forOwn';
-import indexOf from 'lodash/_baseIndexOf';
+import indexOf from 'lodash/indexOf';
```
aearly: Why did you change this?

aearly:
LGTM2.

If there's a tiny DAG library out there that's easy to integrate, maybe.  This code has the advantage of already working with what we've defined to be a "graph", rather than having to construct a graph to verify that it's acyclic.

lib/auto.js
b-paul: @aearly I wrote my `checkForDeadlocks` assuming I had the lodash/JavaScript standard semantics. Got test failures because I hadn't included start index.

Thought I was better off importing the public API method, in case someone else were to come and repeat my mistake at another time

megawac: We weren't importing indexof to avoid including the lodash internal
toInteger implementation

b-paul:
@megawac @aearly I see your concerns about `indexOf`. Thanks for the explanation; I should have guessed it was that way for a reason. I've pushed a commit that reverts the change and fixes the invocation to match

megawac:
Good stuff! thanks

PR #1142, opened 2016-05-04
Make once and onlyOnce exception safe

### LINKED ISSUE #1106: Make once and only_once exception safe
At the moment the following two functions do not behave as expected when an exception occurs:

```
    function only_once(fn) {
        return function() {
            if (fn === null) throw new Error("Callback was already called.");
            fn.apply(this, arguments);
            fn = null;
        };
    }

    function _once(fn) {
        return function() {
            if (fn === null) return;
            fn.apply(this, arguments);
            fn = null;
        };
    }
```

The problem occurs because **fn** is set to **null** too late. So, for example, if an exception were to happen when **fn.apply** is called, the function ptr will not get set to **null** and the callback will potentially get called again. I observed this behaviour when running **async.map** - the final function was called multiple times after an error occurred.

Here's some sample code that illustrates the issue:

```
var async = require('async');
var arrayOfThings = [1, 2];
async.map(arrayOfThings, function (thing, mapNext) {
    try {
        // Do some useful code
        mapNext(thing); // Pretend error occurred
    } catch (thrownErr) {
        mapNext(thrownErr);
    }
}, function (err) {
    console.log('map error: ' + err);
    no_such_function(); // This causes an exception that will cause this function to get called again.
    done();
});
```

The fix is simple:

```
    function only_once(fn) {
        return function() {
            if (fn === null) throw new Error("Callback was already called.");
            var callFn = fn;
            fn = null;
            callFn.apply(this, arguments);
        };
    }

    function _once(fn) {
        return function() {
            if (fn === null) return;
            var callFn = fn;
            fn = null;
            callFn.apply(this, arguments);
        };
    }
```

Cheers,
Neil

ezubarev:
Resolves #1106


megawac:
Mind making a test case and leaving a comment. Otherwise :+1:

aearly:
Yes, a test would be great.

ezubarev:
Done. Anything else?

megawac:
thanks!

PR #1146, opened 2016-05-06
Convert the rest of tests to mocha

ezubarev:
The remaining tests were ported, nodeunit was completely removed. Have a look. Resolves #849
 @megawac, @aearly


mocha_test/timeout.js
```diff
@@ -1,5 +1,5 @@
 var async = require('../lib');
-var expect = require('chai').expect;
+var assert = require('assert');
```
megawac: why this change? Won't this cause errors to be less easy to debug?

ezubarev: Sure, you're right. I'll fix it.

mocha_test/asyncify.js
```diff
@@ -0,0 +1,114 @@
+var async = require('../lib');
+var assert = require('assert');
```
megawac: same here, use `chai.expect` instead of assert please

ezubarev:
Assert in wrapped function could not be handled which led the test result to false positive. Fixed in 210de70.

ezubarev:
@megawac, I replaced assert with expect in tests on asyncify. In other cases as I can see it's usage will not lead to less transparency in debugging. However I don't mind if you want to make use `chai.expect` somewhere else.

megawac:
I'm sure any confusing remaining ones will be fixes post merge. Thanks a bunch @ezubarev

ezubarev:
You're welcome.

PR #1148, opened 2016-05-07
Fix #1147 (bugs preventing use of singleton arrays in auto tasks)

### LINKED ISSUE #1147: Bug in auto/autoInject for tasks of the form `[ <function> ]`
In `.auto()`, tasks with no dependencies may be defined either by a function or an array containing a single function.
The second form actually gives an error cause in [this line](https://github.com/miguel76/async/blob/master/lib/auto.js#L131) the array get enclosed in another array.
This is not crucial `.auto()` itself, but a bit limiting when using `.autoInject()`, cause the array form is an alternative to using explicit parameters, like in the example below.
```javascript
async.autoInject({
  'input-string': [async.apply(fs.readFile, inputFilename, 'utf8')],
  input: ['input-string', async.asyncify(JSON.parse)],
  ...
}, callback); 
```

miguel76:


aearly:
Fixed by e235125

Issue #1001, opened 2016-01-06
Ability to change retry() params from the task?


eexit:
Hello,

I would like to know if it'd be possible to be able to update the `times` argument of the `retry()` from the `task`?
For instance, I would like to be able to increase the `interval` value as the attempt count increases or trim down the attempt count to `0` when a particular error occurred (in order to avoid unnecessary doomed attempts).

Example:

```javascript
var opts = {
    times: 3,
    interval: 500
};

async.retry(opts, function (callback) {
    someCall(function(err) {
        if (err) {
            switch (err.code) {
                case 'Code1':
                    opts.interval += 500; // Increases the interval for next attempt
                    return callback(err);
                case 'Code2':
                    opts.times = 0; // Avoid other attempts
                    return callback(err);
                default:
                    return callback(err);
            }
        }

        callback();
    });
}, function (err) {
    // Some code
});
```

Any idea or workaround?

Thanks,

aearly:
Unfortunately, there's no way to change the retry options after the fact right now -- we copy the options to a new object internally.

To exit early, you could callback with a null error, and some sort of sentinel object you handle specially, but it's kind of a hack.

I'd love to think of a clean way to support customizing the times and interval -- perhaps a function that is called for each iteration to get the times/interval.  This would more easily enable things like exponential backoff and early exits.

```js
function getNextInterval(prevInterval, remaningTimes, lastError) {
  if (prevInterval == null) {
    // initial iteration
    return {
      interval: 100,
      times: 5
    };   
  }
  if (lastError && lastError.code === "foo") {
    // exit early
    return {
      times: 0
    };
  }
  // exponential backoff
  return {
    interval: lastInterval * 2,
    times: times - 1
  };
}

async.times(getNextInterval, worker, done);
```

eexit:
Thanks @aearly! I'll try this eventually and see how it behaves...

murdockcrc:
Could we implement the ```interval``` parameter to support an array of ints? These ints will be used when the ```attempts``` array gets built with the timeout functions that provide the execution delay between retries. I know it does not solve the original request (to change the parameters on the fly), but I am working under the assumption that you don't necessarily need to decide the retry backoff times at runtime.

Obviously, if the passed parameter for ```interval``` is not an array but an Int, then we use that as the timeout for all the retries.

ecasilla:
+1

aearly:
[...]

PR #1161, opened 2016-05-18
Allow custom retry interval

aianus:
Allow interval to be specified as a function(retryCount) : int so users can have exponential backoffs, etc.

#1001

### DIFF
--- a/lib/retry.js
+++ b/lib/retry.js
@@ -19,7 +19,8 @@ import constant from 'lodash/constant';
  * * `times` - The number of attempts to make before giving up.  The default
  *   is `5`.
  * * `interval` - The time to wait between retries, in milliseconds.  The
- *   default is `0`.
+ *   default is `0`. The interval may also be specified as a function of the
+ *   retry count.
  * * If `opts` is a number, the number specifies the number of times to retry,
  *   with the default interval of `0`.
  * @param {Function} task - A function which receives two arguments: (1) a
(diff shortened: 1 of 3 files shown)


aianus:
Needed exponential backoff for a project and figured I'd kick my implementation upstream in case anyone finds it useful.

lib/retry.js
```diff
@@ -129,4 +133,10 @@ export default function retry(times, task, callback) {
             }, interval);
         };
     }
+
+    function constantIntervalFunc(interval) {
```
megawac: can you `import constant from 'lodash/constant'` instead

lib/retry.js
```diff
@@ -95,13 +99,13 @@ export default function retry(times, task, callback) {
         throw new Error("Invalid arguments for async.retry");
     }
 
-
     var attempts = [];
-    while (opts.times) {
-        var isFinalAttempt = !(opts.times -= 1);
+    for (var i = 1; i < opts.times + 1; i++) {
+        var isFinalAttempt = (i == opts.times);
```
megawac: why the change to this loop?

lib/retry.js
```diff
@@ -64,16 +64,20 @@ export default function retry(times, task, callback) {
...
     function parseTimes(acc, t) {
         if (typeof t === 'object') {
             acc.times = +t.times || DEFAULT_TIMES;
-            acc.interval = +t.interval || DEFAULT_INTERVAL;
+
+            if (typeof t.interval === 'function') {
+                acc.intervalFunc = t.interval;
+            } else {
+                acc.intervalFunc = constantIntervalFunc(+t.interval || DEFAULT_INTERVAL);
+            }
```
megawac: ternary this guy

lib/retry.js
aianus: I needed to pass the retryCount to the interval function but the loop was counting backwards and destroying opts.times.

aearly:
This feature also needs to be mentioned in the docs -- also perhaps with an example on how to do exponential back-off, or use a pre-baked array of intervals.

lib/retry.js
```diff
@@ -64,16 +65,19 @@ export default function retry(times, task, callback) {
...
-        interval: DEFAULT_INTERVAL
+        intervalFunc: constant(DEFAULT_INTERVAL)
     };
 
     function parseTimes(acc, t) {
         if (typeof t === 'object') {
             acc.times = +t.times || DEFAULT_TIMES;
-            acc.interval = +t.interval || DEFAULT_INTERVAL;
+
+            acc.intervalFunc = (typeof t.interval === 'function' ?
```
megawac: extraneous bracket.

```js
acc.intervalFunc = typeof t.interval === 'function' ?
                t.interval :
                constant(+t.interval || DEFAULT_INTERVAL);
```

Otherwise looks good to me and +1

megawac:
Thanks!

eexit:
Thanks!

PR #1170, opened 2016-05-29
Add optional error handler to queues

jnordberg:
Allowing you to implement things like error logging and task retries more easily, like so:

```javascript
var q = async.queue(function(task, callback) {
	console.log('Oops.');
	callback(new Error('I fail every time'));
});

q.error = function(error, task) {
	if (task.retries == null) task.retries = 0;
	task.retries++;
	if (task.retries < 3) {
		q.push(task);
	} else {
		console.log('This isnt working...');
	}
};

q.push({foo: 'bar'});
```

### DIFF
--- a/lib/internal/queue.js
+++ b/lib/internal/queue.js
@@ -61,6 +61,10 @@ export default function queue(worker, concurrency, payload) {
                 });
 
                 task.callback.apply(task, args);
+
+                if (args[0] != null) {
+                    q.error(args[0], task.data);
+                }
             });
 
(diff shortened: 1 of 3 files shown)


aearly:
Looks good.  This feature has been requested a few times before.

Can you add a note to the doc block about the feature?

jnordberg:
Sure, to the README right?

megawac:
@jnordberg to the jsdoc above the function preferably. The readme docs are on their way out

PR #1171, opened 2016-05-29
Fix Makefile build on OS X

jnordberg:
Now works on OS X (tested 10.11.4), sed -E should work on linux, not tested..


aearly:
Works on Linux. :+1: 

Thanks for helping sort out all the Unixen.

PR #1172, opened 2016-05-30
Use nexttick instead of setImmediate

Kikobeats:
Basically this change is based on @isaacs reflexion on this post:

http://blog.izs.me/post/59142742143/designing-apis-for-asynchrony

> This uses Node’s synthetic deferral function: process.nextTick, which runs the supplied callback at the end of the current run-to-completion. You can also use setImmediate, but that’s slightly slower. (Yes, yes, it’s named badly; setImmediate is slightly less “immediate” and more “next” than nextTick, but this is an accident of history we cannot correct without a time machine.)

Now `internal/setImmediate` implementation is sort based in best optimal solutions first.

### DIFF
--- a/lib/internal/setImmediate.js
+++ b/lib/internal/setImmediate.js
@@ -4,10 +4,10 @@ import rest from 'lodash/rest';
 var _setImmediate = typeof setImmediate === 'function' && setImmediate;
 
 var _defer;
-if (_setImmediate) {
-    _defer = _setImmediate;
-} else if (typeof process === 'object' && typeof process.nextTick === 'function') {
+if (typeof process === 'object' && typeof process.nextTick === 'function') {
     _defer = process.nextTick;
+} else if (_setImmediate) {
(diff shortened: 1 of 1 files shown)


ex1st:
@Kikobeats nexttick runs before any additional I/O events (including timers) fire in subsequent ticks of the event loop (node api docs), setImmediate - after. This is not the same.

megawac:
:-1: we recommend `setImmediate` for intensive data processing tasks as the event loop will continue before the next iteration is called. This allows some processing with async to not overwhelm the thread. Changing this could mess with peoples code for little benefit. If users want this, call `async.nextTick` instead.

megawac:
I would be open to changing the implementation in https://github.com/caolan/async/blob/master/lib/nextTick.js though @Kikobeats

Kikobeats:
@megawac I agree with your words. The implementation of this PR could be nice for `internal/nextTick`, more consistent.

PR #1173, opened 2016-05-31
Consistent nextTick implementation

Kikobeats:


aearly:
Cool, this is how I had intended to implement `nextTick` when we added support for additional args, but didnt have time to factor out the logic that applies args.

On the other hand, I'm wondering how valuable this change is.  Async wraps `nextTick` and `setImmediate` primarily for polyfill-type purposes.  In the browser you're going to mostly be getting `setTimeout(fn, 0)`, and in Node, primarily `setImmediate`.  The differences between `nextTick` and `setImmediate` are so minor, that those who need `nextTick` will be calling `process.nextTick` directly.  There's also that "recursive nextTick" warning in some versions of node that using setImmediate sidesteps entirely.

Kikobeats:
In addition browserify creates a polyfill of `process.nextTick` 😄 .

The point of this PR is be consistent with the difference between nextTick and setImediate, and I feel that it is specially relevant in the backend side. Because they are not the same, delegate in async to infer what to use looks weird. IMHO It's a programmer responsability. Including recursive patterns.

Kikobeats:
Also in this PR, a little problem with git and CRLF 😢

aearly:
Also a couple conflicts -- no more need for `use strict` in every file.

Kikobeats:
@aearly what do you mean, I should remove `use strict` of the top of the files?

Kikobeats:
I fix the problem with CLR creating and merging a separate branch. Now I don't know how to fix the failed test

Kikobeats:
Fixed, green tests 👍

Issue #1175, opened 2016-05-31
Cleanup package managers.


jdalton:
I noticed async has a lot of package manager scaffolding which could be removed or simplified.
For example component, jam, spmjs, volo could be dropped as they're either no longer maintained or have minuscule usage. The bower.json could be simplified as well. If you all are up for a PR, I'll send one over.

aearly:
Sounds good.  Seems only Bower is the only package manager still kickin' aside from NPM.  Not too familiar with `bower.json1 properties -- what could we remove?

jdalton:
Moved to #1176.

PR #1176, opened 2016-05-31
Reduce package manager scaffolding.

jdalton:
This PR tackles #1175.


support/xyz.sh
```diff
@@ -160,9 +158,6 @@ inc() {
 
 inc package.json
 
-run "$dir/sync-package-managers.js"
-run "git add --force *.json"
```
aearly: Would we still need this line to stage changes to the `package.json`?

aearly:
I love PRs that remove 150 lines :smile:

support/xyz.sh
jdalton: Fixed.

megawac:
+++

PR #1177, opened 2016-05-31
Mapping over Objects

### LINKED ISSUE #1157: Map should only produce array; add mapValues and mapKeys
There was another issue/pr? for this I made sometime last year but currently map can produce either an object or array because of this line:

https://github.com/caolan/async/blob/master/lib/internal/map.js#L11

I would prefer we follow underscore/lodash lead and produce only arrays from map and then add `mapValues` and `mapKeys` for mapping objects (#1133).

This is one of the last changes I want to get into v2. Thoughts @aearly

aearly:
WIP. Closes #1157.

I might add `mapKeys` and family later, but I can't contrive an example use-case where you'd need to map the keys of an object asynchronously.

### DIFF
--- a/lib/internal/map.js
+++ b/lib/internal/map.js
@@ -1,15 +1,23 @@
 import isArrayLike from 'lodash/isArrayLike';
 import getIterator from './getIterator';
 import noop from 'lodash/noop';
+import okeys from 'lodash/keys';
+import indexOf from 'lodash/_baseIndexOf';
 import once from './once';
 
 export default function _asyncMap(eachfn, arr, iteratee, callback) {
     callback = once(callback || noop);
     arr = arr || [];
-    var results = isArrayLike(arr) || getIterator(arr) ? [] : {};
+    var results = [];
+    var keys;
+    if (!isArrayLike(arr) && !getIterator(arr)) {
+        keys = okeys(arr);
+    }
+
     eachfn(arr, function (value, index, callback) {
         iteratee(value, function (err, v) {
-            results[index] = v;
+            var idx = keys ? indexOf(keys, index, 0) : index;
+            results[idx] = v;
             callback(err);
         });
     }, function (err) {
--- a/lib/internal/map.js
+++ b/lib/internal/map.js
@@ -1,23 +1,16 @@
-import isArrayLike from 'lodash/isArrayLike';
-import getIterator from './getIterator';
 import noop from 'lodash/noop';
-import okeys from 'lodash/keys';
-import indexOf from 'lodash/_baseIndexOf';
 import once from './once';
 
 export default function _asyncMap(eachfn, arr, iteratee, callback) {
     callback = once(callback || noop);
     arr = arr || [];
     var results = [];
-    var keys;
-    if (!isArrayLike(arr) && !getIterator(arr)) {
-        keys = okeys(arr);
-    }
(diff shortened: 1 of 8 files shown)


mocha_test/mapValues.js
```diff
@@ -0,0 +1,92 @@
...
+            var running = 0;
+            var concurrency = {
+                a: 2,
+                b: 2,
+                c: 1
+            };
+            async.mapValuesLimit(obj, 2, function (val, key, next) {
+                running++;
+                async.setImmediate(function () {
+                    expect(running).to.equal(concurrency[key]);
```
aearly: BTW, I'm experimenting with this way of verifying concurrency.  It should be more reliable than asserting a concurrency after some timeout value.  Could help with some of our flaky tests.

lib/internal/map.js
```diff
@@ -1,15 +1,23 @@
...
+    var results = [];
+    var keys;
+    if (!isArrayLike(arr) && !getIterator(arr)) {
+        keys = okeys(arr);
+    }
+
     eachfn(arr, function (value, index, callback) {
         iteratee(value, function (err, v) {
-            results[index] = v;
+            var idx = keys ? indexOf(keys, index, 0) : index;
```
megawac: this shouldn't be necessary. Just keep track of by creating and incrementing a numeric index variable (`idx` in the eachfn

```js
var idx = 0;
eachfn(arr, function(...) {
   idx += 1;
   iteratee(...
```

mocha_test/mapValues.js
megawac: Cool, let me know if you want `testingbot` keys to try running in selenium

lib/internal/map.js
aearly: You're right.  Fixed in 611a442

mocha_test/mapValues.js
aearly: We'd have to refactor other tests to be like this style, tracking concurrency and verifying things finish in a certain order, but it might be worth it :sweat:

mocha_test/mapValues.js
```diff
@@ -0,0 +1,92 @@
+var async = require('../lib');
+var expect = require('chai').expect;
+var assert = require('assert');
+
+describe('mapValues', function () {
+    var obj = {a: 1, b: 2, c: 3};
```
megawac: Worth testing with an array as well?

aearly: What would the expected behavior be?

megawac: `{0: <val>, 1: <val>, 2: <val>}`

I would also be fine with leaving this behaviour implied

megawac:
LGTM /cc me when you feel done with this PR and I'll review again+merge

aearly:
It's ready to merge as-is.  I could add `mapKeys`, but as I said, I'm struggling coming up with a real-world-ish example where you'd want to iterate and update the keys of an object asynchronously.

RobertWHurst:
I thanks think of plenty. For one, an object containing urls.
I'm not sure why you decided to remove this.

```javascript
async.map({
  resource1: 'http://resource1.url',
  resource2: 'http://resource1.url'
}, (url, cb) => fetchResource(url, cb), (err, results) => { ... });
```

PR #1182, opened 2016-06-08
Switch to ESLint

### LINKED ISSUE #1111: Stop using JSCS
JSCS is now EOL https://medium.com/@markelog/jscs-end-of-the-line-bc9bf0b3fdb2#.u8kkgu506

aearly:
Closes #1111


lib/race.js
```diff
@@ -19,6 +19,7 @@ import once from './internal/once';
  * @param {Function} callback - A callback to run once any of the functions have
  * completed. This function gets an error or result from the first function that
  * completed. Invoked with (err, result).
+ * @returns undefined
```
aearly: FYI, there's an annoying feature of the ESLint JSDoc validator that makes you document the return value if there's a `return` statement

aearly:
@megawac Can you check my work here?

megawac:
Wow @aearly I think these are the most rigorous linting rules I've ever seen. I'll have to review this again when I'm less tired

aearly:
Nah, it's very similar to what we had before.  The only difference is the JSDoc validator is more strict.  It also caught more indentation discrepancies.

.eslintrc
```diff
@@ -0,0 +1,49 @@
...
+    "sourceType": "module"
+  },
+  "globals": {
+    "self": true,
+    "define": true,
+    "describe": true,
+    "context": true,
+    "it": true,
+    "before": true,
+    "after": true
```
megawac: seperate testing globals from main config?

ex1st: @megawac it would be easier:

```
"env": {
  ....
  "mocha": true
},
```

aearly: Ah, good call, didn't know about the "mocha" env.

PR #1189, opened 2016-06-15
add callbacks to quick examples

freeall:
I got confused for a little bit when I read the quick examples and saw that there was no callback in each of the `.parallel()` and `.series()` functions. So I thought others might also get confused, and added the `callback` parameter to the quick examples.


megawac:
Can you also add these to `parallel.js` and `series.js`

freeall:
Thanks for the merge

PR #1190, opened 2016-06-18
Allow use es6 object method shorthand in autoInject

ivanseidel:
This is a patch to fix the current `parseParams` method, in order to work with the new ES6 Method Shorthand type: http://es6-features.org/#MethodProperties, and also the `default` value to work (it's being stripped in parse)

Parser needed to be modified in order to work with:
* defaults. Eg.: `function (dependency = [], cb){ ... }`
* Method shorthand. Eg.: `autoInject({ methodName (someDependency, cb){}, ... }

Also, the validation of `autoInject task functions require explicit parameters` had to be removed, since shorthand methods do not show up with `length > 0`.

This was done because I see a great advantage in using the method shorthand with defaults, to improve readability. Take this example:

```
async.autoInject({
  forgetTo (callback){
     callback(true);
  },
  dont: (forgetTo, callback) => {
     callback();
  },
  so (dont = 'letMe', forgetTo, callback){
     callback(dont);
  },
}, function (err, results){
  // so: 'letMe
  // dont: undefined
  // forgetTo: true
})
```

### DIFF
--- a/lib/autoInject.js
+++ b/lib/autoInject.js
@@ -125,7 +125,6 @@ export default function autoInject(tasks, callback) {
             if (taskFn.length === 0 && params.length === 0) {
                 throw new Error("autoInject task functions require explicit parameters.");
             }
-            console.log(taskFn.length, params)
 
             params.pop();
 
--- a/lib/autoInject.js
+++ b/lib/autoInject.js
@@ -5,10 +5,19 @@ import clone from 'lodash/_copyArray';
 import isArray from 'lodash/isArray';
 import trim from 'lodash/trim';
 
-var argsRegex = /^(function[^\(]*)?\(?\s*([^\)=]*)/m;
+var FN_ARGS = /^(function)?\s*[^\(]*\(\s*([^\)]*)\)/m;
+var FN_ARG_SPLIT = /,/;
+var FN_ARG = /(=.+)?(\s*)$/;
+var STRIP_COMMENTS = /((\/\/.*$)|(\/\*[\s\S]*?\*\/))/mg;
 
 function parseParams(func) {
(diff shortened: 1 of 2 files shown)


lib/autoInject.js
```diff
@@ -108,8 +117,6 @@ export default function autoInject(tasks, callback) {
             taskFn = params.pop();
 
             newTasks[key] = params.concat(params.length > 0 ? newTask : taskFn);
-        } else if (taskFn.length === 0) {
-            throw new Error("autoInject task functions require explicit parameters.");
```
aearly: I dont like this side effect.  It will catch a really common pitfall where people pass bound (or otherwise wrapped) functions to `autoInject`.  Could you change the conditional to be more like `taskFn.length === 0 && hasNoDefaultParams`?

ivanseidel: Just fixed it @aearly, check it out

aearly:
Cool, this is awesome.

For future reference, I think this is the maximum complexity we want to support for functions passed to `autoInject`.  I don't want to support destructuring since that will require a parser, and I'm not keen on reimplementing part of Babel. :)

ivanseidel:
The only thing I would change, is some way to cache the functions... Parsing/`Regexing` can cost a bit if done multiple times. Maybe `Generating` an async flow and then running it? like:
```
var flow = async.autoInject({
  a: (cb) => {},
  b: (a, cb) => {},
});

flow((err, result) => {
  // Result of operation
});

// And you can run it again, without re-parsing methods and etc.
flow((err, result) => {
  // Result of operation
});
```

Anyway, thanks for allowing me to contribute! =)

PR #1193, opened 2016-06-21
Add `files` property to package.json

xswordsx:
This will allow npm to publish only the `dist/` folder reducing the size of the module.


aearly:
Sorry, this isn't needed because we create a special `build` folder in the publish process with exactly what needs to be published.

PR #1197, opened 2016-06-23
Callback can be called twice in asyncify when using promises.

raydog:
Code to reproduce the bug:
```javascript

const i_return_a_promise = () => Promise.resolve("Hello");

const fn = async.asyncify(i_return_a_promise);

var _count = 0;
fn((err) => {
  console.log("CB CALL #%d", ++_count, err);
  throw new Error("D:");
});
```
This code will print out:
```
CB CALL #1 null
CB CALL #2 [Error: D:]
```
This is because `.then` on a promise will catch errors, and return a rejected promise, which would then be caught by the `.catch` code, calling the callback twice. This PR fixes that behavior by handling both the resolve case and the rejection case in the `.then` handler.

### DIFF
--- a/mocha_test/asyncify.js
+++ b/mocha_test/asyncify.js
@@ -74,6 +74,13 @@ describe('asyncify', function(){
             'rsvp'
         ];
 
+        // Both Bluebird and native promises emit these events. We handle it because Bluebird
+        // will report these rejections to stderr if we don't, which is a great feature for
+        // normal cases, but not here, since we expect unhandled rejections:
+        process.on('unhandledRejection', function () {
+            // Ignore.
+        });
+
         names.forEach(function(name) {
             describe(name, function() {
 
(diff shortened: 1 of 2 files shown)


megawac:
Good catch, 100% agree this is the correct behaviour

mocha_test/asyncify.js
```diff
@@ -74,6 +74,13 @@ describe('asyncify', function(){
             'rsvp'
         ];
 
+        // Both Bluebird and native promises emit these events. We handle it because Bluebird
+        // will report these rejections to stderr if we don't, which is a great feature for
+        // normal cases, but not here, since we expect unhandled rejections:
+        process.on('unhandledRejection', function () {
+            // Ignore.
+        });
```
megawac: Just realized, can you remove this after the test completes? `process.off` or only bind it with `process.once`

Please make a followup pr

PR #1205, opened 2016-06-30
Implment queues using DLLs

### LINKED ISSUE #756: async.queue performance suffers greatly with large queues
While performance testing a library that uses async.queue, I noticed that the performance of async.queue degrades massively as the number of items in the queue increases. After a certain point, V8 spends so much time shuffling memory around that the performance of everything else slows to a crawl.

The culprit appears to be how the current implementation calls `shift` on the tasks array every time an item is processed. I prototyped an alternate implementation that maintains an index into the array and performance is vastly improved with larger queue sizes. For the purposes of this testing, the entire async.queue implementation was copied and modified to form a new async.indexedQueue function so side-by-side comparison could be easily done.

https://gist.github.com/bbrowning/b7355b731eb025608228

megawac:
A linked list is a more optimal method to keep track of a stack of tasks as we can end up mutating the `tasks` very frequently. The previous solution could have been further optimized to avoid several splices (i.e. doing 1 splice instead of n in several scenarios), though a linked list solution should generally be more performant.

/cc @aearly 

Fixes #756

### DIFF
--- a/lib/internal/DoublyLinkedList.js
+++ b/lib/internal/DoublyLinkedList.js
@@ -1,0 +1,64 @@
+// Simple doubly linked list (https://en.wikipedia.org/wiki/Doubly_linked_list) implementation
+// used for queues. This implementation assumes that the node provided by the user can be modified
+// to adjust the next and last properties. We implement only the minimal functionality
+// for queue support.
+export default function DLL() {
+    this.head = this.tail = null;
+    this.length = 0;
+}
+
+function setInitial(dll, node) {
+    dll.length = 1;
+    dll.head = dll.tail = node;
+}
+
+DLL.prototype.removeLink = function(node) {
+    if (!node) return node;
+    if (node.prev) node.prev.next = node.next;
+    else this.head = node.next
+    if (node.next) node.next.prev = node.prev;
+    else this.tail = node.prev;
+
+    node.prev = node.next = null;
+    this.length -= 1;
+    return node;
+}
+
+DLL.prototype.empty = DLL;
+
+DLL.prototype.insertAfter = function(node, newNode) {
+    newNode.prev = node;
+    newNode.next = node.next;
+    if (node.next) node.next.prev = newNode;
+    else this.tail = newNode;
+    node.next = newNode;
+    this.length += 1;
+}
+
+DLL.prototype.insertBefore = function(node, newNode) {
+    newNode.prev = node.prev;
+    newNode.next = node;
+    if (node.prev) node.prev.next = newNode;
+    else this.head = newNode;
+    node.prev = newNode;
+    this.length += 1;
+}
+
+DLL.prototype.unshift = function(node) {
+    if (this.head) this.insertBefore(this.head, node);
+    else setInitial(this, node);
+};
+
+DLL.prototype.push = function(node) {
+    if (this.tail) this.insertAfter(this.tail, node);
+    else setInitial(this, node);
+};
+
+
+DLL.prototype.shift = function() {
+    return this.removeLink(this.head);
+};
+
+DLL.prototype.pop = function() {
+    return this.removeLink(this.tail);
+};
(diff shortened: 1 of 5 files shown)


megawac:
also /cc @mikestopcontinues @bbrowning @zeluis9
If any of you have a second can you verify this solves your performance issues?

lib/priorityQueue.js
```diff
@@ -62,25 +48,26 @@ export default function(worker, concurrency) {
...
         arrayEach(data, function(task) {
             var item = {
                 data: task,
                 priority: priority,
                 callback: typeof callback === 'function' ? callback : noop
             };
 
-            q.tasks.splice(_binarySearch(q.tasks, item, _compareTasks) + 1, 0, item);
-
-            setImmediate(q.process);
```
megawac: note this change, moved process outside of the loop. I think this is the correct behaviour

mikestopcontinues:
I'll be happy to, but it might have to wait until the weekend. Deadlines! :)

lib/internal/DoublyLinkedList.js
```diff
@@ -0,0 +1,62 @@
...
+    node.prev = node.next = null;
+    this.length -= 1;
+    return node;
+}
+
+DLL.prototype.empty = DLL;
+
+DLL.prototype.insertAfter = function(node, newNode) {
+    newNode.prev = node;
+    newNode.next = node.next;
```
aearly: Will this add `prev` and `next` properties to items people push to queues?  Perhaps the list should store `{prev, next, data}` objects, where `data` points to the original object.

aearly:
Also, what do our benchmarks look like with these changes?   (Just realized I haven't tried to run those since modularization.)

lib/internal/DoublyLinkedList.js
megawac: no, push wraps the item in a object, `{data, callback}` for queue and `{data, callback, priority}` for priority queue

megawac:
```
./perf/benchmark.js -g queue
Latest tag is  v2.0.0-rc.6
Comparing v2.0.0-rc.6 with current on Node v6.2.0
queue(1000) v2.0.0-rc.6 x 298 ops/sec ±5.46% (28 runs sampled), 3.35ms per run
queue(1000) current x 316 ops/sec ±5.05% (29 runs sampled), 3.16ms per run
current is faster
queue(30000) v2.0.0-rc.6 x 8.06 ops/sec ±8.54% (15 runs sampled), 124ms per run
queue(30000) current x 8.95 ops/sec ±4.06% (17 runs sampled), 112ms per run
current is faster
queue(100000) v2.0.0-rc.6 x 0.14 ops/sec ±0.00% (1 run sampled), 7.12e+3ms per run
queue(100000) current x 2.54 ops/sec ±9.01% (6 runs sampled), 394ms per run
current is faster
queue(200000) v2.0.0-rc.6 x 0.04 ops/sec ±0.00% (1 run sampled), 2.70e+4ms per run
queue(200000) current x 1.22 ops/sec ±31.59% (3 runs sampled), 817ms per run
current is faster
current faster overall (1330ms total vs. 34200ms total)
current won more benchmarks (4 vs. 0)
```

aearly:
Damn, thats impressive.  Looks like with small queues it will be slower, but at those sizes, the list overhead is negligible anyway.

aearly:
Also, I realize this is a subtle breaking change -- the `tasks` array was exposed (but undocumented).  Some people might be relying on that array for custom splicing.

megawac:
Ye, that's why I changed the variable name and want to get this in v2.
Anyone relying on tasks being an array will get a reference error

megawac:
Also based on C's intuition-fu I'd expect this to be faster in all cases
(except some edge priority queue ones) =d
On Jun 30, 2016 6:38 PM, "Graeme Yeates" <megawac@gmail.com> wrote:

Ye, that's why I changed the variable name and want to get this in v2.
Anyone relying on tasks being an array will get a reference error

mikestopcontinues:
For the record, this totally fixes my issue. Thanks @megawac. You rock!

megawac:
good to hear :tada: @mikestopcontinues thanks for getting back

PR #1207, opened 2016-07-01
ReflectAll allow to accept object of functions

andalm:
I made this PR for solve this issue #1195 

```
let tasks = {
     one: function(callback) {
         setTimeout(function() {
             callback(null, 'one');
         }, 200);
     },
     two: function(callback) {
         callback('two');
     },
     three: function(callback) {
         setTimeout(function() {
             callback(null, 'three');
         }, 100);
     }
 };

 async.parallel(async.reflectAll(tasks),
 // optional callback
 function(err, results) {
     // values
     // results.one.value = 'one'
     // results.two.error = 'two'
     // results.three.value = 'three'
 });
```

If you have any feedback just let me know, thanks.

PR #1208, opened 2016-07-01
ReflectAll, use lodash functions instead native functions

andalm:
Fixes of #1207 
@megawac's suggestions

### DIFF
--- a/lib/reflectAll.js
+++ b/lib/reflectAll.js
@@ -1,6 +1,6 @@
 import reflect from './reflect';
 import isArray from 'lodash/isArray';
-import map from 'lodash/map';
+import _arrayMap from 'lodash/_arrayMap';
 import forOwn from 'lodash/forOwn';
 
 /**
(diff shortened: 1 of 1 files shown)


lib/reflectAll.js
```diff
@@ -1,5 +1,7 @@
 import reflect from './reflect';
 import isArray from 'lodash/isArray';
+import map from 'lodash/map';
```
megawac: arrayMap

lib/reflectAll.js
```diff
@@ -1,5 +1,7 @@
...
+import forIn from 'lodash/forIn';
```
megawac: forOwn

lib/reflectAll.js
andalm: @megawac I can't replace this function because lodash v4.8.0 arrayMap doesn't exists.

megawac: @andalm see https://github.com/caolan/async/blob/4931114cd6844b3d7c5e7e9d4e6d3b54d0ad95ab/lib/internal/filter.js#L1 for reference

andalm: Done

PR #1210, opened 2016-07-01
Improve doc site layout

ivanseidel:
This fixes a few issues with responsiveness, and improves the search a bar a bit.
screen shot 2016-07-01 at 3 09 42 pm


support/jsdoc/theme/static/styles/jsdoc-default.css
```diff
@@ -127,12 +143,32 @@ section, h1 {
...
+.navbar-nav>li, .navbar-nav {
+  float: left !important;
+}
+
+.navbar-nav.navbar-right:last-child {
+  margin-right: -15px !important;
+}
+
+.navbar-right {
+  float: right!important;
```
aearly: Is it possible to accomplish what you want without using `!important` everywhere?

support/jsdoc/jsdoc-custom.js
```diff
@@ -99,7 +99,7 @@ $(function initSearchBar() {
             location.href = host + 'docs.html#.' + suggestion;
         } else {
             var $el = document.getElementById('.' + suggestion);
-            $('#main').animate({ scrollTop: $el.offsetTop }, 500);
+            $('#main').animate({ scrollTop: $el.offsetTop - 60 }, 500);
```
aearly: Why is the `-60` required?

support/jsdoc/theme/static/styles/jsdoc-default.css
ivanseidel: Those `!important` are to override the default bootstrap behavior..
[http://stackoverflow.com/a/23536146/1234504](http://stackoverflow.com/a/23536146/1234504)

support/jsdoc/jsdoc-custom.js
ivanseidel: I changed the `h1` top padding to `0` in order to accomplish a left blue vertical bar in each header, but. Now, the spacing between contents is by setting a top `margin`, and `margins` are not used to calculate the `(x,y)` position of the element. Because of that, the scroll was not giving much "space" before (it was perfectly aligned with the top nav bar..)

support/jsdoc/theme/static/styles/jsdoc-default.css
aearly: Ah, i see.  It might not be required, since this stylesheet is loaded after bootstrap.

I'm also considering ditching Bootstrap and just doing the CSS ourselves.  We're fighting it in a couple places, and all we use it for is this navbar.

ivanseidel: I guess that the layout can be done, and then re-written without bootstrap. I can help with that later, but would be good to set the 'target' layout now, so that later, it is only migrated...

ivanseidel:
Hey @aearly, did this got published to the site yet? I didn't see the changes there http://caolan.github.io/async/

aearly:
I haven't had a chance to publish yet.

PR #1212, opened 2016-07-02
Implement callIf

dmh43:
Recently I've been working on a project where this functionality was useful. Sometimes, I'd like to conditionally call an asynchronous function, but always call some callback afterwards. I've implemented `callIf` here. I'm still unsure about the name.

### DIFF
--- a/lib/callIf.js
+++ b/lib/callIf.js
@@ -1,0 +1,36 @@
+/**
+ * Calls the provided function if the conditional is truthy. A callback is given
+ * which takes an error, optional return value, and the value of the conditional argument.
+ *
+ * @name callIf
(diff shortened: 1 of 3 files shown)


megawac:
:-1: I don't think this is better than plain JS

```js
var cond = 1 === 2;

cond ? fn(cb) : cb();
```

aearly:
Yeah, historically, Async has been against adding conditional logic helpers -- plain JS syntax ends up being simpler.

I did play around with a [`provided`](https://www.npmjs.com/package/acomb#provided) method in a helper library.  It's very similar to what you have here.  However, in practice, I never actually used it for anything.

PR #1213, opened 2016-07-03
Avoid including lodash iteratee in builds

megawac:
`forOwn` and `each` include `baseIteratee` which in turn includes all the `matches` logic and `isEqual` logic. This kills a lot of unused lodash functionality. Likely related to those space savers jdd found with using webpack

/cc @aearly


aearly:
Awesome!

Yeah, I'd love if there was a `lodash-rollup-plugin` or something similar, so we could get the same features without having to switch everything to Webpack.  This should be a major improvement, nonetheless.

PR #1214, opened 2016-07-03
Optimize queue processing

megawac:
Applies some JIT  optimizations for arguments and avoids iterating the entire worker set. Then just uses the idle helper because thats what its there for :)

### DIFF
--- a/lib/internal/queue.js
+++ b/lib/internal/queue.js
@@ -1,5 +1,4 @@
 import arrayEach from 'lodash/_arrayEach';
-import arraySome from 'lodash/_arraySome';
 import isArray from 'lodash/isArray';
 import noop from 'lodash/noop';
 import rest from 'lodash/rest';
@@ -51,10 +50,10 @@ export default function queue(worker, concurrency, payload) {
             workers -= 1;
 
             arrayEach(tasks, function (task) {
-                arraySome(workersList, function (worker, index) {
+                arrayEach(workersList, function (worker, index) {
                     if (worker === task) {
                         workersList.splice(index, 1);
-                        return true;
+                        return false;
                     }
                 });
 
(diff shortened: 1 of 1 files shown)


lib/internal/queue.js
```diff
@@ -45,16 +47,14 @@ export default function queue(worker, concurrency, payload) {
...
-        return function(){
+        return rest(function(args){
             workers -= 1;
 
-            var removed = false;
-            var args = arguments;
             arrayEach(tasks, function (task) {
-                arrayEach(workersList, function (worker, index) {
-                    if (worker === task && !removed) {
+                arraySome(workersList, function (worker, index) {
```
megawac: Note, this is a slight change in behaviour as more than 1 task can be removed per process. I think the fix should be only doing the rest of this functions processing in the outer arrayEach as we currently call tasks not in the workerList. I think we need to discuss this and identify some unit tests for this case

aearly: I don't think there's any harm in doing it this way -- there's no way this `_next` call can be preempted by another task completing.

Also -- does `arrayEach` support the early return?  e.g. `return false`.

megawac: currently a task can be called multiple times because we call all tasks I think.

Also yes on arrayEach, why?

aearly: I'm pretty sure we only execute each task once, because we remove them as they are processed.  I could be missing some detail though.

I think using `some` for short circuiting is a bit hacky.  It also means we don't have to import `arraySome`.

megawac: The thing is we don't remove them on each run because we check if `removed
= true` so we only remove one task at a time while possibly running multiple

aearly:
After closer inspection, this should be fine.  The inner loop is effectively `workersList.remove(task)`, where each `task` is unique.

PR #1218, opened 2016-07-05
Prevent arguments leak (compose)

ex1st:
For details: [Optimization killers](https://github.com/petkaantonov/bluebird/wiki/Optimization-killers#32-leaking-arguments)


megawac:
use `args.reverse()`

PR #1221, opened 2016-07-08
Run unit tests on build

megawac:
/cc @aearly how would you run this on the modules? right now this is only testing `dist/async`


aearly:
`support/build.test.js`  is where we test the build right now.  It just does a quick smoke test to verify things seem okay in all the formats we support (single files, monolithic, ES6).  If you want to beef up the testing of `dist/async.js` I'd do it there.

Otherwise, when working on the code, you'd have to run the  build before your ran the mocha tests, a bit of an annoyance.

megawac:
As far as I could tell it was only being run on `dist/async`, I'd like the entire test suite to be run on all distributions of async (which is what this is going for). Any objection?

aearly:
I think it's overkill to run the tests on all the various ways we distribute it. (Single file, individual modules, single ES6 modules, ES6 index, etc..)  I think if we verify the build puts thing in the right place, that's good enough, if one full suite of the tests passes.

Take a second look at `support/build.test.js` -- it makes sure things gets put in their right places.  I'd rather not overcomplicate our general mocha tests.

PR #1224, opened 2016-07-10
Fix test function arguments for whilst/until/during

### LINKED ISSUE #1217: async.during iteratee callback do not support optional parameter
I am using v2.0.0 on client side.

In async.during example, if we pass a optional value in callback of iterate, javascript show error of `callback is not a function` in truthy functon.

```
var count = 0;

async.during(
    function (callback) {
        console.warn(arguments);
        return callback(null, count < 5);
    },
    function (callback) {
        count++;
        setTimeout(function(){
          callback(null, count);
        }, 1000);
    },
    function (err, res) {
        console.info(res); // 5 seconds have passed
    }
);
```

Is there anything I am doing wrong? I guess this is also mentioned in documentation that *`callback` will be passed an error and any arguments passed to the final `fn`'s callback*

aearly:
Closes #1217.

cc @megawac

### DIFF
--- a/lib/doWhilst.js
+++ b/lib/doWhilst.js
@@ -1,6 +1,8 @@
 import noop from 'lodash/noop';
 import rest from 'lodash/rest';
 
+import onlyOnce from './internal/onlyOnce';
+
 /**
  * The post-check version of [`whilst`]{@link module:ControlFlow.whilst}. To reflect the difference in
  * the order of operations, the arguments `test` and `iteratee` are switched.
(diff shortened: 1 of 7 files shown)


lib/doWhilst.js
```diff
@@ -12,20 +13,23 @@ import whilst from './whilst';
...
+ * `callback` will be passed an error and any arguments passed to the final
+ * `iteratee`'s callback. Invoked with (err, [results]);
  */
-export default function doWhilst(fn, test, callback) {
-    var calls = 0;
-    whilst(function() {
-        return ++calls <= 1 || test.apply(this, arguments);
-    }, fn, callback);
+export default function doWhilst(iteratee, test, callback) {
+    callback = callback || noop;
```
megawac: lets `once` wrap this

megawac:
:+1: I like this resolution. Another option would be to change the `test` expected definition to

```
test(callback, arrayofarguments)
```

aearly:
I think this is the best solution.  Either you have a test function that has a different signature the first time its called, you create an async function that doesn't have a callback as its last arg, or you avoid the situation entirely.

megawac:
Can you update changelog

aearly:
88dab8c :D

PR #1227, opened 2016-07-11
Fixes #1226, make retry interval asynchronous

### LINKED ISSUE #1226: retry: unable to log "Retrying in x minutes"
When using `retry` or `retryable`, I would like to log that the retry will happen in _x_ milliseconds/seconds/minutes.

For instance,
```
[Jul 11, 2016 14:25:37] This API key has reached their API limit for the hour. Please wait before making another call.
[Jul 11, 2016 14:25:37] Retrying in 5 minutes.
```

If I try to log this in the interval function, it does not work because the interval function precomputes the intervals and pops them into another async method.

The task callback function doesn't make sense because that happens _after_ the interval has elapsed and I don't know anything about the retry interval (nor should I). 

Maybe there should be another callback that can be invoked prior to waiting on the interval time period?

2.0.0-rc6

Node 6.2.1

megawac:
Also makes the retry code much simpler (and probably faster)

### DIFF
--- a/lib/retry.js
+++ b/lib/retry.js
@@ -1,4 +1,3 @@
-import series from './series';
 import noop from 'lodash/noop';
 import constant from 'lodash/constant';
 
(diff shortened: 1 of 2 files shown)


aearly:
:+1:

Very nice.

Issue #1229, opened 2016-07-12
PriorityQueue execution order


suguru03:
**What version of async are you using?**
v2.0.0-rc.6, master

**Which environment did the issue occur in (Node version/browser version)**
Node

**What did you do? Please include a minimal reproducable case illustrating issue.**
https://github.com/caolan/async/blob/master/mocha_test/queue.js#L405-L431
If I changed `queue` to `priorityQueue`, the order is reversed.

```js
var call_order = [];
var q = async.priorityQueue(function (task, callback) {
    if (task.isLongRunning) {
        q.pause();
        setTimeout(function () {
            call_order.push(task.id);
            q.resume();
            callback();
        }, 50);
    }
    else {
        call_order.push(task.id);
        setTimeout(callback, 10);
    }
}, 10);

q.push({ id: 1, isLongRunning: true});
q.push({ id: 2 });
q.push({ id: 3 });
q.push({ id: 4 });
q.push({ id: 5 });

q.drain = function () {
    expect(call_order).to.eql([1, 2, 3, 4, 5]); // [5, 4, 3, 2, 1]
    done();
};
```

**What did you expect to happen?**
I think first task should be executed at first if priority wasn't defined.

**What was the actual result?**
`[ 5, 4, 3, 2, 1 ]`

aearly:
@megawac probabaly related to the DLL changes.

PR #1230, opened 2016-07-12
fix execution order

suguru03:
Related #1229
If `priority` wasn't defined, `undefined >= undefined` is false.
What do you think about default value?
I'm not sure which is better 0 or 1.

### DIFF
--- a/lib/priorityQueue.js
+++ b/lib/priorityQueue.js
@@ -51,6 +51,7 @@ export default function(worker, concurrency) {
             });
         }
 
+        priority = priority || 0;
         var nextNode = q._tasks.head;
         while (nextNode && priority >= nextNode.priority) {
             nextNode = nextNode.next;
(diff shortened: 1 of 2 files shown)


megawac:
I kind of consider this case an error and I'm pretty sure if you were to implement this in another language we'd error when provided no priority.

I guess the thing we should decide is whether `priority` undefined means highest priority or lowest priority.

aearly:
Yeah, I like this.  "no priority" === "priority 0"  seems logical.

megawac:
That depends on whether we want to acknowledge negative priorities as being possibly a thing. I think this requires more discussion @aearly. 

Like what should be the order in the case of

```js
        q.push({ id: 1, isLongRunning: true});
        q.push({ id: 2 }, 2);
        q.push({ id: 3 }, 3);
        q.push({ id: 4 }, -1);
        q.push({ id: 5 });
```

Currently its 

```
4 1 5 2 3
```

I'd argue it should be

```
1 5 4 2 3
```

PR #1239, opened 2016-07-12
Remove async.iterator

### LINKED ISSUE #1237: To kill async.iterator or not to kill
As far as I can tell, `async.iterator` is not used internally despite being documented as an exposed internal method. As mentioned in #1220, I think it's time we leave this method in the dust in favour of ES2015 iterator protocol (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Iterators_and_Generators) which natively supports arrays

megawac:
Closes #1237


aearly:
:+1:

aearly:
There's been some flakiness in the `retry` tests.  I hope its just travis being weird.

PR #1241, opened 2016-07-14
Use mocha tag for non-browser tests, fix asyncify promisified browser tests

### LINKED ISSUE #1235: Browser tests appear to be running with non-browser code
I attempted to refactor the tests to use mocha "tags" to prevent non-browser tests from running in the browser (i.e. instead of using `isBrowser` or `process.browser`).  See https://github.com/tekwiz/async/commit/176b947043500767818c3b67396dfd9fb26260af.  However, when I did this, the browser tests failed to run:

```
START:
12 07 2016 12:51:02.153:INFO [framework.browserify]: bundle built
12 07 2016 12:51:02.169:INFO [karma]: Karma v0.13.22 server started at http://localhost:9876/
12 07 2016 12:51:02.176:INFO [launcher]: Starting browser Firefox
12 07 2016 12:51:04.238:INFO [Firefox 45.0.0 (Mac OS X 10.11.0)]: Connected on socket /#IAMzwTa1Tv27JMNsAAAA with id 45738836
Firefox 45.0.0 (Mac OS X 10.11.0) ERROR
  Error: Cannot find module 'native-promise-only'
  at /var/folders/yl/pg2l6p6j5nb9twhsp8ymqktc0000gn/T/87c1e0334e1cc54f87ea66cc4683870f.browserify:1 <- node_modules/browser-pack/_prelude.js:1:0

Finished in 0.425 secs / 0 secs

SUMMARY:
✔ 0 tests completed
```

I was able to narrow down the problem to `mocha_test/asyncify.js` -- it occurs when the asyncify - promisified suite is not loaded.  My theory is that browserify or browser-pack is attempting to load native-promise-only because it is referenced but the other is not loading it because it's never actually run.

I'm not really sure how go about debugging this, so any pointers would be appreciated.

tekwiz:
Removes browser checks in tests.  Fixes asyncify promisified to run in browser (fixes #1235)

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -12,6 +12,13 @@ module.exports = function(config) {
         browserify: {
             debug: true,
             transform: ['babelify']
+        },
+
+        client: {
+            mocha: {
+                grep: '@nodeonly',
+                invert: true
+            }
         }
     });
 };
--- a/mocha_test/asyncify.js
+++ b/mocha_test/asyncify.js
@@ -1,7 +1,6 @@
 var async = require('../lib');
 var assert = require('assert');
 var expect = require('chai').expect;
-var isBrowser = require('./support/is_browser');
 
 describe('asyncify', function(){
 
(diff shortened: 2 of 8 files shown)


megawac:
Nice :exclamation:

PR #1245, opened 2016-07-20
Minor cleanups to eachOfLimit

megawac:
Although I made a couple micro perf optimizations (avoiding a couple extra function calls) there are no real changes to performance.

I did find another minor optimization was to aggressively set the `iterateeCallback` to null when done (in which case we could also kill the done variable)


aearly:
LGTM.  Even if it doesn't affect performance, at least we can get rid of that eslint loopfunc exception.

megawac:
Actually I ran the bench marks wrong (forgot to compile first). This is a slight optimization (~5%). Going to test the aggressive optimization now

PR #1247, opened 2016-07-21
Make iterator more prone to JIT optimizations

megawac:
This just separates the types of iterators out into atomic functions instead of handling all types in one spot


aearly:
:+1:

PR #1257, opened 2016-07-31
Enhance waterfall, series, parallel and parallelLimit

wood1986:


megawac:
whats is the motivation behind this? I think I'm :-1:

aearly:
Yeah, this also breaks the callback-last convention we just fixed in all our methods.  In most cases you wouldn't need to integrate this with Async -- you could just zip up the results into an object at the end.

wood1986:
- Lets start from `waterfall`. The main purpose of the zip is for not only the end but also each task. Suppose I am at the 3rd task, I want to use the result which is calculated at the 1st task. You will say I can define a variable outside the `waterfall` and then zip it before calling `callback`.
A tedious comes out
```
var zip = {};
async.waterfall([
  (callback) => {
    mongo.collection("async").findOne({}, {}, (err, doc) => { // tedious callback
      zip["1"] = doc;
      callback(null, zip);
    });
  },
  (args, callback) => {
    mongo.collection("async").findOne({}, {}, (err, doc) => { // tedious callback
      zip["2"] = doc;
      callback(null, zip);
    });
  }
], (err, callback) => {
});
```

versus

```
var zip = {};
async.waterfall([
  (callback) => {
    mongo.collection("async").findOne({}, {}, callback);
  },
  (args, callback) => {
    mongo.collection("async").findOne({}, {}, callback);
  }
], (err, callback) => {
},
["1", "2"]);
```
Does it make sense?

- It is parallel*` and `series`'s turn. Due to my proposal of a new `waterfall` and `parallel*` and `series` similarity, I apply the same design into them. That also gives `series` an advantage that the existing one cannot. With the new changes, the tasks will be executed in order and the result will be in object format

- It is break in the middle and last callback tradition's turn. Theoretically, `done` is another kind of  `callback`. What I can see is if you insist on this rule ("last callback tradition"), finally you must violate  
the rule ("err first") in order to have a break in the middle for `waterfall` because you repurpose the err. Also I do not think you will introduce an variable or a function between `err` and `callback` to this. What I can see that `class` will be your final choice and that's why some threads mention this feature should belong to 3.0.0

```
async(
  {
    "type": "waterfall",
    "tasks": [ () => { }, () => { }, () => { } ]
    "callback": () =>
  }
)();
```

PR #1258, opened 2016-07-31
Enhance waterfall, series, parallel and parallelLimit (no break in the middle feature)

wood1986:
Please see my [latest reply](https://github.com/caolan/async/pull/1257)

### DIFF
--- a/lib/internal/parallel.js
+++ b/lib/internal/parallel.js
@@ -1,5 +1,6 @@
 import noop from 'lodash/noop';
 import isArrayLike from 'lodash/isArrayLike';
+import isPlainObject from 'lodash/isPlainObject';
 import rest from 'lodash/_baseRest';
 
 export default function _parallel(eachfn, tasks, callback) {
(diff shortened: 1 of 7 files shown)


wood1986:
Exact same function signature

```
async.waterfall([
  {"1": (callback) => { mongo.collection("async").findOne({}, {}, callback); } },
  {"2": (args, callback) => { mongo.collection("async").findOne({}, {}, callback); } }
], (err, callback) => {
});
```

megawac:
I understand your use case but I don't think this belongs in async core. I'd suggest making a tiny wrapper method for this instead

hargasinski:
The core already does something similar to your use case already, at least for the `parallel` and `series` cases. If an object is passed, the object's keys as used as the keys for the `results` object, allowing you to use custom names for the keys. See the examples for [parallel](http://caolan.github.io/async/docs.html#.parallel) and [series](http://caolan.github.io/async/docs.html#.series).

wood1986:
I definitely know `parallel` and `series` have the similar thing but it is not perfect. Other than that, parallel and series are not my target. My target is `waterfall`. Meanwhile, I find my changes can be applied into `series` to fix the limitation of the execution order with returning a result in object format.

aearly:
If you need keyed results in a waterfall-like sceneario, `auto` is more up that alley.

wood1986:
@aearly Although I thumb up, the cons of `auto` is I have to type a lot.

PR #1261, opened 2016-08-02
add filter option to retry() and retryable(). PR for #1256.

### LINKED ISSUE #1256: error filter for retry and retryable
Hello,

I did q quick search and didn't find similar issue... It would be useful to have `retry` and `retryable` accept an optional `errorFilter` sync function that will only allow the retry to continue if the function returns `true`. for example something like:

```
var retryOpts = {
  errorFilter: function (err) {
    if(err.code === 11) return true; // only retry for error code 11
    return false;
  }
};

async.retry(retryOpts, apiMethod, function(err, result) {
  // We know that we retried on error code 11, with other retry options being the defaults. 
  // if we got any other error we immediately bailed to here
  if(err) console.log(err); 
  console.log(result);
});
```

This could be done in backwards compatible manner. Opinions? Would this be something accepted? I could probably try and do a PR sometime next week if so.

Thanks,

Bojan

bojand:
* add filter option to `retry()` and `retryable()` to allow for error filtering and control of retry flow. Resolves #1256.
* Added comment regarding the new option. Do I need to run the script to update the generated documentation?
* The new option is called `errorFilter`.
* Added tests. Lint and all existing tests pass when I run with `npm test`.

Fixes #1256 - megawac

### DIFF
--- a/lib/retry.js
+++ b/lib/retry.js
@@ -19,6 +19,10 @@ import constant from 'lodash/constant';
  * * `interval` - The time to wait between retries, in milliseconds.  The
  *   default is `0`. The interval may also be specified as a function of the
  *   retry count (see example).
+ * * `filter` - Synchronous function that is invoked on erroneous result with the
+ *   the error. If it returns `true` the retry attempts will continue, if the
+ *   function returns `false` the retry flow is aborting with the current
+ *   attempt's error and result being returned to the final callback.
  * * If `opts` is a number, the number specifies the number of times to retry,
  *   with the default interval of `0`.
  * @param {Function} task - A function which receives two arguments: (1) a
@@ -62,6 +66,16 @@ import constant from 'lodash/constant';
  *     // do something with the result
  * });
  *
+ * // try calling apiMethod only when error condition satisfies, all other
+ * // errors will abort the retry control flow and return to final callback
+ * async.retry({
+ *   filter: function(err) {
+ *     return err.message === 'Temporary error'; // only retry on a specific error
+ *   }
+ * }, apiMethod, function(err, result) {
+ *     // do something with the result
+ * });
+ *
  * // It can also be embedded within other control flow functions to retry
  * // individual methods that are not as reliable, like this:
  * async.auto({
@@ -70,6 +84,7 @@ import constant from 'lodash/constant';
  * }, function(err, results) {
  *     // do something with the results
  * });
+ *
  */
 export default function retry(opts, task, callback) {
     var DEFAULT_TIMES = 5;
@@ -87,13 +102,16 @@ export default function retry(opts, task, callback) {
             acc.intervalFunc = typeof t.interval === 'function' ?
                 t.interval :
                 constant(+t.interval || DEFAULT_INTERVAL);
+
+            if(typeof t.filter === 'function') {
+                acc.filter = t.filter;
+            }
         } else if (typeof t === 'number' || typeof t === 'string') {
             acc.times = +t || DEFAULT_TIMES;
         } else {
             throw new Error("Invalid arguments for async.retry");
         }
     }
-
 
     if (arguments.length < 3 && typeof opts === 'function') {
         callback = task || noop;
@@ -111,7 +129,16 @@ export default function retry(opts, task, callback) {
     function retryAttempt() {
         task(function(err) {
             if (err && attempt++ < options.times) {
-                setTimeout(retryAttempt, options.intervalFunc(attempt));
+                var proceed = true;
+                if(options.filter) {
+                    proceed = options.filter(err);
+                }
+
+                if(proceed) {
+                    setTimeout(retryAttempt, options.intervalFunc(attempt));
+                } else {
+                    callback.apply(null, arguments);
+                }
             } else {
                 callback.apply(null, arguments);
             }
--- a/lib/retry.js
+++ b/lib/retry.js
@@ -19,11 +19,11 @@ import constant from 'lodash/constant';
  * * `interval` - The time to wait between retries, in milliseconds.  The
  *   default is `0`. The interval may also be specified as a function of the
  *   retry count (see example).
- * * `continueOperation` - An optional synchronous function that is invoked on
- *   erroneous result with the the error. If it returns `true` the retry attempts
- *   will continue, if the function returns `false` the retry flow is aborted
- *   with the current attempt's error and result being returned to the final
- *   callback. Invoked with (err).
+ * * `errorFilter` - An optional synchronous function that is invoked on
+ *   erroneous result. If it returns `true` the retry attempts will continue;
+ *   if the function returns `false` the retry flow is aborted with the current
+ *   attempt's error and result being returned to the final callback.
+ *   Invoked with (err, result).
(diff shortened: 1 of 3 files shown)


lib/retry.js
```diff
@@ -19,6 +19,10 @@ import constant from 'lodash/constant';
  * * `interval` - The time to wait between retries, in milliseconds.  The
  *   default is `0`. The interval may also be specified as a function of the
  *   retry count (see example).
+ * * `filter` - Synchronous function that is invoked on erroneous result with the
```
megawac: I don't think filter is the right term for this. Perhaps `continueOperation` or something? /cc @hargasinski @aearly 
Also this should be documented as optional

lib/retry.js
```diff
@@ -87,14 +102,17 @@ export default function retry(opts, task, callback) {
             acc.intervalFunc = typeof t.interval === 'function' ?
                 t.interval :
                 constant(+t.interval || DEFAULT_INTERVAL);
+
+            if(typeof t.filter === 'function') {
+                acc.filter = t.filter;
+            }
```
megawac: no need for the conditional just set it here and check if it is a function when you execute

lib/retry.js
```diff
@@ -111,7 +129,16 @@ export default function retry(opts, task, callback) {
...
-                setTimeout(retryAttempt, options.intervalFunc(attempt));
+                var proceed = true;
+                if(options.filter) {
+                    proceed = options.filter(err);
+                }
+
+                if(proceed) {
+                    setTimeout(retryAttempt, options.intervalFunc(attempt));
+                } else {
+                    callback.apply(null, arguments);
```
megawac: should it not be called with err?

lib/retry.js
```diff
@@ -111,7 +129,16 @@ export default function retry(opts, task, callback) {
     function retryAttempt() {
         task(function(err) {
             if (err && attempt++ < options.times) {
-                setTimeout(retryAttempt, options.intervalFunc(attempt));
+                var proceed = true;
```
megawac: var proceed = typeof options.filter != 'function' || options.filter(err);

lib/retry.js
hargasinski: :+1: for `continueOperation`. I'd also be open to `errorFilter` from the issue, 

Also, you don't need to run the script to generate the documentation. Just make sure the tests pass.

bojand: I wasn't fan of `filter` either. Most of the do / while type of functions within async use `test`... maybe something similar to that? `continueTest`?

lib/retry.js
bojand: `callback` is applied with `null` for context / this, and `arguments` for the actual arguments, so the error should be passed through. The new tests cover this. This is consistent with the scenario where we've exhausted all attempts and still had an error possibly.

bojand:
Thank you for the feedback, it's appreciated. I will address all the issues once we settle on the new parameter name.

lib/retry.js
bojand: Missed the comment somehow. Thanks for the feedback. W will change it to `continueOperation`.

aearly: `errorFilter`/`errorTest`/`shouldError`/`onError`/`shouldContinueOnError`

I think `continueOperation` is a bit misleading, because it's not something you do when you continue, it's a test/callback to see if you should continue or not.

Cache invalidation and naming things....

lib/retry.js
```diff
@@ -19,6 +19,11 @@ import constant from 'lodash/constant';
  * * `interval` - The time to wait between retries, in milliseconds.  The
  *   default is `0`. The interval may also be specified as a function of the
  *   retry count (see example).
+ * * `continueOperation` - An optional synchronous function that is invoked on
+ *   erroneous result with the the error. If it returns `true` the retry attempts
+ *   will continue, if the function returns `false` the retry flow is aborted
+ *   with the current attempt's error and result being returned to the final
+ *   callback. Invoked with (err).
```
megawac: isn't it invoked with `(error, ...any other arguments)`

lib/retry.js
```diff
@@ -111,7 +128,13 @@ export default function retry(opts, task, callback) {
...
-                setTimeout(retryAttempt, options.intervalFunc(attempt));
+                var proceed = typeof options.continueOperation != 'function' ||
+                    options.continueOperation(err);
+                if(proceed) {
+                    setTimeout(retryAttempt, options.intervalFunc(attempt));
+                } else {
+                    callback.apply(null, arguments);
+                }
             } else {
                 callback.apply(null, arguments);
```
megawac: can you try to merge these two else conditions

lib/retry.js
bojand: I like `errorFilter` or `errorTest`. Another option I suppose is `retryTest` ?

mocha_test/retryable.js
```diff
@@ -21,6 +21,30 @@ describe('retryable', function () {
...
+        });
+
+        retryableTask(42, function (err) {
+            expect(err).to.equal('fail');
+            expect(calls).to.equal(3);
+            done();
+        });
+
+        setTimeout(function () {
+        }, 15);
```
megawac: qu'est-ce que?

mocha_test/retry.js
```diff
@@ -156,5 +156,105 @@ describe("retry", function () {
...
+            callback(err, erroredResult + callCount);
+        }
+        function errorTest(err) {
+            return err && err !== special;
+        }
+        var start = new Date().getTime();
+        async.retry({ interval: interval, errorFilter: errorTest }, fn, function(err, result){
+            var now = new Date().getTime();
+            var duration = now - start;
+            assert(duration >= (interval * (specialCount - 1)),  'did not include interval');
```
megawac: this is a bit fragile of a test

lib/retry.js
```diff
@@ -19,6 +19,11 @@ import constant from 'lodash/constant';
  * * `interval` - The time to wait between retries, in milliseconds.  The
  *   default is `0`. The interval may also be specified as a function of the
  *   retry count (see example).
+ * * `errorFilter` - An optional synchronous function that is invoked on
+ *   erroneous result. If it returns `true` the retry attempts will continue;
+ *   if the function returns `false` the retry flow is aborted with the current
+ *   attempt's error and result being returned to the final callback.
+ *   Invoked with (err, result).
```
megawac: It's currently invoked with just err

mocha_test/retry.js
bojand: hi can you elaborate? Do you mean the specific assertion? I based it on existing test. Just checks that the appropriate time has passed?

megawac: its probably fine, we have just seen issues lately with some tests erring out (especially in CI) due to slight timing issues

mocha_test/retryable.js
megawac: What is this timeout for?

bojand: I had used an existing test which had it; and I didn't notice it. Not sure why it's there to be honest. In the latest commit I removed it from the existing test and the new one. Tests pass locally for me, and in CI it seems.

megawac:
Thanks!

PR #1263, opened 2016-08-04
Documentation typo

ieatfood:
"Invoked with" doubled.


hargasinski:
Thanks!

hargasinski:
@megawac how often do we update the gh-pages? Should it be after every PR or just for releases/ major changes?

aearly:
The docs get automatically published with every release, but if you want to push out a small fix like this (and there are no new features waiting to be released), all you have to do is `make publish-docs` as a contributor.

hargasinski:
Oh okay, thank you! This is the first fix to the docs since the last update, so I pushed it.

Issue #1266, opened 2016-08-10
Segmentation Fault: 11, large async.queue size


rtran9:
**What version of async are you using?**
`v4.4.7`

**Which environment did the issue occur in (Node version/browser version)**
`async@2.0.1`

**What did you do? Please include a minimal reproducible case illustrating issue.**
```
var async = require("async");

var count = 0;
var queue = async.queue(function(item, callback) {
    count += 1;
    console.log(count);
    callback(null);
});

queue.drain = function () {
    console.log("done");
};

var data = new Array(20000);
queue.push(data);
```

**What was the actual result?**

Node throws this error `Segmentation fault: 11` after popping some of the items from the queue.
```
...
13440
13441
13442
13443
13444
13445
13446
13447
13448
13449
13450
13451
Segmentation fault: 11
```

To get this to run in the first place without a `Maximum call stack size exceeded` error, I increased the stack size: `node --stack-size=32000`.

Can anyone explain to me whats going on? It works up until a certain array size.

hargasinski:
See #1265 for a little more info.

Essentially, `queue` expects the `worker` to call `callback` asynchronously. As mentioned in that issue, you can use [async.ensureAsync](http://caolan.github.io/async/docs.html#.ensureAsync) or [async.setImmediate](http://caolan.github.io/async/docs.html#.setImmediate) for your case to ensure that `callback` is called asynchronously.

Using `async.setImmediate`, your code would change to:
```js
var async = require("async");

var count = 0;
var queue = async.queue(function(item, callback) {
    count += 1;
    console.log(count);
    async.setImmediate(callback, null);
});

// rest of your code... 
```  

or using `async.ensureAsync`:
```js
var async = require("async");

var count = 0;
var queue = async.queue(async.ensureAsync(function(item, callback) {
    count += 1;
    console.log(count);
    async.setImmediate(callback, null);
}));

// rest of your code...
```

In terms of a little more detailed explanation of what's happening to answer your question. The `call stack` basically manages memory for a program; it can only hold a maximum number of functions. Every time you call a function synchronously, you add it to the stack. When you call another function from that function, you again add it to the top of the stack and so on. The functions don't get removed until they return (it's a little more complicated then that, but that's the basic idea). If the stack runs out of space, an error is thrown.

[...]

Issue #1265, opened 2016-08-10
eachSeries with a nested async call results in maximum call stack size exceeded


mymainmanbrown:
**What version of async are you using?**

2.0.1

**Which environment did the issue occur in (Node version/browser version)**

Node 6.2.1
NPM 3.9.3

**What did you do? Please include a minimal reproducable case illustrating issue.**

```
let counter = 2000;

async.eachSeries(
  new Array(counter),
  (item, cb) => {
    async.parallel([
      (pCb) => {
        console.log(counter--);
        pCb();
      },
    ], cb);
  },
  () => console.log('done')
);
```

**What did you expect to happen?**

I expected each of the iterations to occur, which would result in the last two console.log's being '1' and 'done'.

**What was the actual result?**

After fewer than 1000 iterations, I would receive this error:

RangeError: Maximum call stack size exceeded

When I reverted back to async 1.5.2, the above code worked as expected.

clickclickonsal:
+1

jhyman2:
+1

megawac:
The reason this worked in async 1.5.2 is we ensured the callback was always called asynchronously in that version. In version 2 we removed this guard (see https://github.com/caolan/async/blob/master/intro.md#synchronous-iteration-functions)

So in your case to fix the script you can either use `async.ensureAsync(cb)` or `async.setImmediate(cb)` to work around the recursion issue:

```js
var counter = 2000;

async.eachSeries(
  new Array(counter),
  (item, cb) => {
    async.parallel([
      (pCb) => {
        console.log(counter--);
        async.setImmediate(pCb);
      },
    ], cb);
  },
  () => console.log('done')
);
```

PR #1267, opened 2016-08-11
prevent stack growth if queue callback is called synchronously

autopulated:
This prevents issues like #1266 and #1265.

By setting a flag which is set only during the execution of the queue callback, the callback can know whether it's being called synchronously. If it is, then it doesn't need to call process(), as it has already decreased the worker count, so the existing [worker loop](https://github.com/caolan/async/compare/master...Coggle:master#diff-4a7415ea0f194634d3f062b247fe15f2L102) will already loop an additional time and re-assign work.

Aside from preventing stack overflow errors, even if the stack is relatively small this change still reduces memory consumption whenever there is a synchronous call to the callback, and should improve performance as a result.

(The only changes are in `lib/internal/queue.js`, I'm not sure why my re-build changed the order of a bunch of things in the re-built dist files)


hargasinski:
We don't rebuild the files in `dist` after each PR, so you're seeing a combination of the last few changes.

@megawac, @aearly this could be a good way of bringing back support for synchronous callbacks async had back in `1.5.2`. Adding a check should have a negligible performance impact, but would expand the use cases for the library. Thoughts?

aearly:
This PR is kind of like automatically wrapping each worker in `ensureAsync`.  We just used a major release to get rid of unnecessary deferrals throughout Async -- I don't want to move back in that direction.

I'm also not convinced that this implementation doesn't need to call `q.process()` after a task completion if the worker was sync -- there's a gap in testing there.  I feel the queue will inadvertently get paused in certain cases.

Synchronous iteration is one of the common pitfalls -- we call that out in the docs.  Just wrap your worker function with `ensuireAsync`, or defer your callback and it will be fine.

hargasinski:
Good points, thanks

autopulated:
I don't think this is like automatically wrapping in ensureAsync – that incurs a performance penalty, whereas this is an optimisation.

Could you clarify what the gap in testing is – If that were covered could this (without the re-built files) be merged?

It doesn't need to call `q.process` again because if the call is synchronous then the main loop of the `q.process` function will inevitably be run again if there's still work to do, as when the callback is called it decreases the workers count.

megawac:
I want to look into this in more detail

syrnick:
Looks like the queue doesn't get paused, because the _next will decrement the number of workers inside the while loop in `process`.  The next iteration of the while loop will do the same work as q.process() inside _next would've done before this change. The only thing it eliminates is at least 3 stack entries (onlyOnce, _next and process) per item in the queue. 

It's a great enhancement, as synchronous iteration is often happening in code that is mostly async, but sometimes could be sync. E.g. say, you have a function that hits the database. That's obviously async. Now you implement an in-memory cache somewhere deep down in that function. The interface has to be async because mostly it'd be hitting the database, but sometimes in-memory cache. You might not have any issues for a while, but when you have high volume and the cache starts working well, you suddenly hit synchronous iteration and blow up with stack size exceeded.

megawac:
The reason I'm hesitant to accept something like this is it is sort of a can of worms. If we start handling synchronous and asynchronous handlers differently in one spot we'll end up having to do it in other places.

This in turn can introduce issues such as stack overflow issues for large iterations in addition to other errors. Although I definitely like the idea of trying to increase the performance of things like queue, I feel like distinguishing between synchronous and asynchronous handlers is the wrong approach. I'll be experimenting with other methods to determine whether process is necessary.

Thanks anyway for the request @syrnick

autopulated:
@megawac FWIW I'm not suggesting that anything more should be "supported" than currently is – only improving the support under the hood in this one case.

Since the javascript language doesn't provide any support for tail-calling or for releasing the memory associated with the function's arguments before returning from it there's unfortunately no other way to achieve this than detecting when the function is called synchronously (I'd be happy to be proved wrong if you find one, but I satisfied myself that there isn't before resorting to this less than elegant approach!)

megawac:
Actually, I think I have a slightly different way of doing this which may be a bit more reasonable (will want thoughts from @aearly and @autopulated )

megawac:
See #1334, it's just a slight twist on what this PR is doing

aearly:
#1334 seems to be doing essentially the same thing, just a bit more cleanly.

One thought after further inspection -- if your worker is synchronous, concurrency is irrelevant.  With this change, you'd basically have infinite concurrency.  Any and all jobs in the queue would be processed on the same tick.

megawac:
> One thought after further inspection -- if your worker is synchronous, concurrency is irrelevant. 

True, if you're using queue such that all functions are either synchronous or asynchronous. I'm under the impression users are mixing asynchronous and synchrnous queue handlers

aearly:
There is a use case for the queue of spreading CPU intensive tasks across multiple ticks.  People would now have to wrap them in `ensureAsync` or defer the callback.

megawac:
Another thing I'm considering is deferring the `q.process` call to the next tick (I'll cc you on a commit in a second)

megawac:
Something like https://github.com/caolan/async/commit/11b699744d480de4c68373755132e91e522d3c90

unfortunately this seems to break unit tests (though I wouldn't expect it to)

PR #1280, opened 2016-09-09
Fixing inconsistency in whilst jsdoc

hairyhenderson:
The docs are a little confusing here, since they refer to `fn`, but the signature names that argument `iteratee`. I'm just clarifying it here...

Signed-off-by: Dave Henderson <dhenderson@gmail.com>


hairyhenderson:
Thanks @aearly!

PR #1282, opened 2016-09-13
Fix misspelled word

platedodev:
colleciton to collection


megawac:
wrong file you want this one https://github.com/caolan/async/blob/master/lib/eachLimit.js

platedodev:
@megawac Oh thanks, change done mate :)

megawac:
Thanks for catching this

PR #1297, opened 2016-10-05
Improve filter by reducing complexity.

CodeMan99:
This change reduces the complexity of `filter` by removing the need to sort the asynchronous result.

Instead, simply map to a temporary array with the truthy state. In the each function callback filter the original array with the truthy state at the same index.

### DIFF
--- a/lib/internal/filter.js
+++ b/lib/internal/filter.js
@@ -1,31 +1,19 @@
-import arrayMap from 'lodash/_arrayMap';
-import property from 'lodash/_baseProperty';
+import filter from 'lodash/_arrayFilter';
 import noop from 'lodash/noop';
 import once from './once';
 
 export default function _filter(eachfn, arr, iteratee, callback) {
     callback = once(callback || noop);
-    var results = [];
+    var truthy = [];
     eachfn(arr, function (x, index, callback) {
         iteratee(x, function (err, v) {
-            if (err) {
-                callback(err);
-            }
-            else {
-                if (v) {
-                    results.push({index: index, value: x});
-                }
-                callback();
-            }
+            truthy[index] = !!v;
+            callback(err);
         });
     }, function (err) {
-        if (err) {
-            callback(err);
-        }
-        else {
-            callback(null, arrayMap(results.sort(function (a, b) {
-                return a.index - b.index;
-            }), property('value')));
-        }
+        if (err) return callback(err);
+        callback(null, filter(arr, function (_, index) {
+            return truthy[index];
+        }));
     });
 }
--- a/lib/internal/filter.js
+++ b/lib/internal/filter.js
@@ -1,19 +1,19 @@
-import filter from 'lodash/_arrayFilter';
+import filter from 'lodash/filter';
 import noop from 'lodash/noop';
 import once from './once';
 
 export default function _filter(eachfn, arr, iteratee, callback) {
     callback = once(callback || noop);
-    var truthy = [];
+    var truthValues = new Array(arr.length);
     eachfn(arr, function (x, index, callback) {
         iteratee(x, function (err, v) {
-            truthy[index] = !!v;
+            truthValues[index] = !!v;
             callback(err);
         });
     }, function (err) {
         if (err) return callback(err);
         callback(null, filter(arr, function (_, index) {
-            return truthy[index];
+            return truthValues[index];
         }));
     });
 }
--- a/mocha_test/filter.js
+++ b/mocha_test/filter.js
@@ -41,6 +41,18 @@ describe("filter", function () {
         });
     });
 
+    it('filter collection', function(done){
+        var a = {a: 3, b: 1, c: 2};
+        async.filter(a, function(x, callback){
+            callback(null, x % 2);
+        }, function(err, results){
+            expect(err).to.equal(null);
+            expect(results).to.eql([3,1]);
+            expect(a).to.eql({a: 3, b: 1, c: 2});
+            done();
+        });
+    });
+
     it('filter error', function(done){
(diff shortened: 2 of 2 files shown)


megawac:
I like it! Can you run `./perf/benchmark.js --grep "filter"` (you'll have to rebase just fixed the benchmarks)

CodeMan99:
Should I be using `"lodash/_arrayFitler"` directly (vs `"lodash/fitler"`)?

megawac:
> Should I be using "lodash/_arrayFilter" directly (vs "lodash/filter")?
Yes, was going to bring that up :)

CodeMan99:
On node v4.6.0 (built from source), the benchmark said that two perfs were faster. Results in [this gist](https://gist.github.com/77b30601072645b73dd9bf90162b4591).

megawac:
@CodeMan99 just to make sure you ran `make` before running perf tests?

CodeMan99:
Pretty sure, I'll run again... yup! I definitely ran a `make` to build.

lib/internal/filter.js
```diff
@@ -1,31 +1,19 @@
-import arrayMap from 'lodash/_arrayMap';
-import property from 'lodash/_baseProperty';
+import filter from 'lodash/_arrayFilter';
 import noop from 'lodash/noop';
 import once from './once';
 
 export default function _filter(eachfn, arr, iteratee, callback) {
     callback = once(callback || noop);
-    var results = [];
+    var truthy = [];
```
megawac: Can we change this variable to to something like `var truthyIndexes` or `filteredIndexes`.

I'd also be curious about the performance of making an array of accepted indexes rather than a truthy list of indexes:

For instance an array might look like `[1, 5, 6]` if we wanted to take those indexes

the final filter array building step would look like

```js
var length = filteredIndexes.length;
var results = new Array(length)
for (var i = 0; i < length; i++) {
   results[i] = arr[filteredIndexes[i]];
}
return results
```

I'd expect this to be faster

CodeMan99: The creation of the pre-filtered indexes is hard (`[1, 5, 6]`). This is why I went with a true/false at every index, the order is easy because indexes don't need to change.

megawac: shouldn't be too hard, you have access to `index` from line 9

just `filteredIndexes.push(index)` no?

megawac: oh right, sorry.

aearly:
:+1:

Performance-wise, this seems to be a bit of a wash, but it does significantly simplify the code.

CodeMan99:
Wrote a basic test using an object as the input. I am assuming order, which is technically wrong, but I expect it to pass everywhere. Feel free to update it.

You were correct that using `lodash/_arrayFilter` will not work with an object, so I just changed back to `lodash/filter`.

I changed "truthy" to "truthValues" to be slightly more clear.

lib/internal/filter.js
```diff
@@ -1,19 +1,19 @@
-import filter from 'lodash/_arrayFilter';
+import filter from 'lodash/filter';
```
CodeMan99: It iterates on the `arr` argument, which is documented as a "collection" type. The test I wrote would fail when using `lodash/_arrayFilter`.

lib/internal/filter.js
```diff
@@ -1,31 +1,19 @@
...
-            if (err) {
-                callback(err);
-            }
-            else {
-                if (v) {
-                    results.push({index: index, value: x});
-                }
-                callback();
-            }
+            truthValues[index] = !!v;
```
megawac: so for objects we're assigning `x[key]` where x is an array. Thus creating a sparse array. I think I'd be more comfortable making `x` an object.

lib/internal/filter.js
```diff
@@ -1,31 +1,19 @@
-import arrayMap from 'lodash/_arrayMap';
-import property from 'lodash/_baseProperty';
+import filter from 'lodash/filter';
```
megawac: the reason we've avoided importing `filter` up until this point is it included a bunch of deep property stuff (to support keys like `a.b['c']` that we do not need in async. I think I'd prefer we assume that `arr` is always an object and use `lodash` object methods instead of lodash collection methods

mocha_test/filter.js
```diff
@@ -41,6 +41,18 @@ describe("filter", function () {
         });
     });
 
+    it('filter collection', function(done){
```
megawac: 💯

lib/internal/filter.js
```diff
@@ -1,31 +1,19 @@
...
-            callback(err);
-        }
-        else {
-            callback(null, arrayMap(results.sort(function (a, b) {
-                return a.index - b.index;
-            }), property('value')));
-        }
+        if (err) return callback(err);
+        callback(null, filter(arr, function (_, index) {
+            return truthValues[index];
```
megawac: either unroll this as a simple loop `keys + for` loop or use a lodash object method and I'll approve

megawac:
I was thinking about this again and started implementing my suggestions (#1320) but now I realize there is another issue we have to consider: `filter` supports iterables which won't be supported by `_.filter`/`_.each` 

:sigh:

CodeMan99:
@megawac Apologies for not progressing this. When trying to implement your suggestions I found it slowed the method by a considerable amount. After that my schedule filled up with many other things. 👷

megawac:
yeah, no worries. I usually try and take over myself when it gets to the nitty gritty stuff like this that doesn't have a clear answer... but like you I've been super busy lately. 

Anyway, I'm not sure what the best path going forward on this PR/#1320

PR #1300, opened 2016-10-08
added scroll-spy behaviour (fixes #1289)

### LINKED ISSUE #1289: improve async documentation with scroll-spy behavior
It would be nice if when scrolling the async documentation page, the method in view would get highlighted on the left.

Something similar to how [express documentation](http://expressjs.com/en/4x/api.html) does it. There, as you scroll down, the methods on the right get highlighted. 

This behavior should be (easily?) achievable with [Bootstrap scrollspy](http://getbootstrap.com/javascript/#scrollspy) and Bootstrap is already being included in the website for documentation anyways...

Just saying...

hargasinski:
I just wanted a second opinion on this PR because I had to change the ids of the method sections in the body for scroll-spy to work. I just had to remove the `.` in front the ids (`.eachSeries` vs `eachSeries` now). This means that potentially old links to methods in the docs will just point to the top of the old page. I added some js that should minimise this, it replaces the hash in the link with the hash with the `.` removed. It should direct users to the correct method.

I think the ids are a result of the way minami generates them. Its something I should've changed initially when creating the docs. Sorry about that.

Also, to get scroll-spy to work, I had to rearrange some of the HTML/CSS, but from what I can see, everything looks the same as before. I have a live demo on my fork: [http://hargasinski.github.io/async](http://hargasinski.github.io/async).

Thanks in advance for the feedback. 

/cc @megawac @aearly @kalmanh

### DIFF
--- a/support/jsdoc/jsdoc-custom.js
+++ b/support/jsdoc/jsdoc-custom.js
@@ -102,4 +102,16 @@ $(function initSearchBar() {
             $('#main').animate({ scrollTop: $el.offsetTop - 60 }, 500);
         }
     });
+
+    function fixOldHash() {
+        var hash = window.location.hash;
+        if (hash) {
+            var hashMatches = hash.match(/^#\.(\w+)$/);
+            if (hashMatches) {
+                window.location.hash = '#'+hashMatches[1];
+            }
+        }
+    }
+
+    fixOldHash();
 });
--- a/support/jsdoc/jsdoc-custom.js
+++ b/support/jsdoc/jsdoc-custom.js
@@ -96,10 +96,10 @@ $(function initSearchBar() {
             location.href = host + suggestion;
         // handle searching from one of the source files or the home page
         } else if (currentPage !== 'docs.html') {
-            location.href = host + 'docs.html#.' + suggestion;
+            location.href = host + 'docs.html#' + suggestion;
         } else {
-            var $el = document.getElementById('.' + suggestion);
-            $('#main').animate({ scrollTop: $el.offsetTop - 60 }, 500);
+            var $el = document.getElementById(suggestion);
+            $('#main-container').animate({ scrollTop: $el.offsetTop - 60 }, 500);
         }
     });
 
--- a/support/jsdoc/jsdoc-fix-html.js
+++ b/support/jsdoc/jsdoc-fix-html.js
@@ -132,9 +132,7 @@ function scrollSpyFix($page, $nav) {
 
     });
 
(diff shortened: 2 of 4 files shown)


megawac:
I'm fine with these changes in foc structure though it seems kind of weird
how much we're fighting the template...

Also can the .each loop be removed on l135 and be just a .addClass?

hargasinski:
Fixed, thanks for catching that! Should I squash the commits or is this good to merge?

The main reason I'm changing the template so much after the docs are generated is because it's a lot quicker to support the `category` tag that way, while keeping all of the main documentation on one page. That being said, I'll start looking into template, and how it can be updated, so we don't need the `jsdoc-fix-html.js` to be as big as it is. We'll still probably need it for things like copying the `dist`, and logo files over to the docs.

megawac:
Works for me in current state (though I haven't seen the actual behaviour
so I'll leave merging to you or Alex

hargasinski:
Thank you! I just had to fix the searching as it was relying on the old ids. Everything appears to be working now.

megawac:
@hargasinski want to tag a release?

https://github.com/caolan/async/compare/v2.0.1...master

hargasinski:
@megawac, sure! Just to confirm, that'll mean updating the changelog, building the dist files and updating the version to `2.1.0`, right? Then creating the actual release on GitHub.

I just want to confirm so I don't mess anything up :weary:. Thanks!

megawac:
Workflow is updating the changelog and then running `make release-[major|minor|patch|prerelease]` I believe

hargasinski:
I should've looked at the `make` file. Thanks, I'll get right on it!

hargasinski:
@megawac I tagged the release. Everything went smoothly updating Github, except for one small linting error I fixed. However, `make release-minor` threw an `npm ERR! you do not have permission to publish "async". Are you logged in as the correct user? : async` error during the `cd build/ && npm publish` step as I don't have permission to publish to npm (I think?, I logged in using `npm login` using hargasinski). You or @aearly will build the files locally, and publish them to npm. The GitHub steps are done.

megawac:
done, had to release a 2.1.1 though

On Wed, Oct 12, 2016 at 2:38 PM, Hubert Argasinski <notifications@github.com

hargasinski:
Thanks!

aearly:
Good work, I like the enhancement!

One thing to think about for the future is how to simplify the docs generation code.  We are fighting the template a lot.

PR #1304, opened 2016-10-13
docs: fix typo in expected result for`mapValues()`

mikermcneil:
This replaces "array" with "object" (looks like it was likely a copy/paste error from when these docs were originally brought over from `.map()`), and rephrases the sentence accordingly.  I also normalized an occurrence of "result" vs. "results" to match.

> *NOTE:* I did not yet go through and do the same to `.mapValuesLimit()` and `.mapValuesSeries()` in this commit.  I'm happy to take care of that in this PR, but it's purely a matter of copy and pasting the relevant comments.

### DIFF
--- a/lib/mapValues.js
+++ b/lib/mapValues.js
@@ -27,7 +27,7 @@ import doLimit from './internal/doLimit';
  * transformed value. Invoked with (value, key, callback).
  * @param {Function} [callback] - A callback which is called when all `iteratee`
  * functions have finished, or an error occurs. `result` is a new object consisting
- * of each key from `obj`, with each transformed result on the right-hand side.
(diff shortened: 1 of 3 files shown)


megawac:
Ah thanks good catch. New wording looks good. Can you comment when you update `mapValuesLimit` and `*Series`

mikermcneil:
@megawac 👍 

I copied the relevant bit across to the others verbatim, with one exception: made a minor tweak to the language from the last time you reviewed:

>  ...with each transformed **value** on the right-hand side...

(I noticed that there was some pre-existing words to that effect in `mapValuesSeries`, so I wanted to emulate that.  It also more clearly captures the meaning, I think, since that's the way the variable for the incoming rvalue passed into the iteratee is labeled)

PR #1306, opened 2016-10-15
Short circuit loops completely for some,every,detect when ready

### LINKED ISSUE #1293: detectSeries breaks when iterating over large arrays
Async 2.0.1, Node v0.10.29. 

detectSeries on a large array causes a stack overflow even if the callbacks are called via setImmediate. Test case:

```

var arr = [];
for (var i = 0; i < 10000; i++) {
        arr.push(i);
}

// this should print 0, then print done, because the detection succeeds on the
// first element of arr. It does, but then it causes a stack overflow
async.detectSeries(arr, function(data, cb) {
        console.log(data);
        setImmediate(cb, null, true);
        // cb(null, true) is definitely expected to cause a stack overflow
        // but setImmediate does not appear to change the situation
}, function(err, result) {
        console.log("done");
});
```

```
0
done
```

```
0
done

.../node_modules/async/dist/async.js:843
        return function () {
                        ^
RangeError: Maximum call stack size exceeded
```

The bug appears to be tied to the size of arr. arr with 1000 items does not trigger the issue on my system.

megawac:
Add the ability to short circuit `eachLimit` and `eachSeries` loops internally. Break loops when value is determined for some, every and detect/find

Fixes #1293

### DIFF
--- a/mocha_test/every.js
+++ b/mocha_test/every.js
@@ -1,5 +1,6 @@
 var async = require('../lib');
 var expect = require('chai').expect;
+var _ = require('lodash');
 
 describe("every", function () {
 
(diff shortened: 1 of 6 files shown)


hargasinski:
Looks good to me

mocha_test/every.js
```diff
@@ -83,6 +84,32 @@ describe("every", function () {
...
+            done();
+        });
+    });
+
+    it('everyLimit doesn\'t cause stack overflow (#1293)', function(done) {
+        var arr = _.range(10000);
+        let calls = 0;
+        async.everyLimit(arr, 100, function(data, cb) {
+            calls += 1;
+            async.setImmediate(_.partial(cb, null, false));
```
aearly: FYI, `setImmediate` will partially apply params for you, no need for `_.partial`

megawac: I thought our implementation didn't (though didn't check :))

PR #1307, opened 2016-10-15
Update dependencies and travis testing

megawac:


megawac:
failing cause eslint apparently depends on node 4

PR #1312, opened 2016-10-23
tiny documentation/comment correction

gilesbowkett:
Fixed typo or stale comment. It implied that `map` takes two callbacks, but it really just takes one.

(Sorry to be the guy who files a whole PR for a typo, I'm sort of a compulsive proofreader.)


gilesbowkett:
sorry, it looks like this fix is wrong.

PR #1317, opened 2016-11-01
make code examples in docs runnable using RunKit

hargasinski:
[RunKit](https://runkit.com/home) essentially creates a sandboxed JS environment to play around with code in the browser. I thought it would be a good way of implementing runnable code examples for the docs. The main advantage is that the environment can be loaded dynamically, so we don't have to slow down the page by loading all 70+ runnable code snippets at once. Lodash already uses it in their [docs](https://lodash.com/docs/4.16.5#chunk).

I have a live demo on my fork: http://hargasinski.github.io/async/docs.html. The [reduce](http://caolan.github.io/async/docs.html#reduce) and [times](http://hargasinski.github.io/async/docs.html#times) docs are probably best for testing it as they don't use `fs`.

The PR isn't ready to be merged yet, as the main issue right now is that a few of our examples use `fs`. I was thinking the best way to handle it would be to change those examples so they don't use `fs`. We could also mock the `fs` functions, similar to how Lodash does it in the [lodash-doc-globals](https://www.npmjs.com/package/lodash-doc-globals) module. I just wanted feedback before moving forward with any of these options. Thanks.

Also, after running a code snippet, there is an `undefined` logged:
screenshot_20161031_222712
That's the `module.exports` value. I don't think there's much I could do about it.

### DIFF
--- a/lib/reduce.js
+++ b/lib/reduce.js
@@ -40,6 +40,7 @@ import once from './internal/once';
  *     });
  * }, function(err, result) {
  *     // result is now equal to the last value of memo, which is 6
+ *     console.log(result);
  * });
  */
 export default function reduce(coll, memo, iteratee, callback) {
--- a/lib/times.js
+++ b/lib/times.js
@@ -31,6 +31,7 @@ import doLimit from './internal/doLimit';
  *     });
  * }, function(err, users) {
  *     // we should now have 5 users
+ *     console.log(users);
  * });
  */
 export default doLimit(timesLimit, Infinity);
--- a/support/jsdoc/jsdoc-custom.js
+++ b/support/jsdoc/jsdoc-custom.js
@@ -103,6 +103,33 @@ $(function initSearchBar() {
         }
     });
 
+    $('.runkit-run-btn').click(function() {
+        var $btn = $(this);
+        var $container = $btn.closest('.runnable-code');
+        $container.children().hide();
+
+        var source = "var async = require('async');\n";
+        var codeTxt = $container.children('pre').text();
+
+        if (/fs\./.test(codeTxt)) {
+            source += "var fs = require('fs');\n\n" + codeTxt;
+        } else {
+            source += '\n'+codeTxt;
+        }
+
+        var notebook = RunKit.createNotebook({
+            // the parent element for the new notebook
+            element: $container.get(0),
+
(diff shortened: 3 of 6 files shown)


megawac:
perhaps ajax call examples instead of fs?

hargasinski:
@megawac good idea. I'm pretty busy the next couple of days, but I'll try to get a couple of the `fs` examples converted to ajax examples soon just to see how it look/work.

aearly:
A lot of our examples aren't designed to be runnable -- they're mostly for illustration purposes.  You'd have to modify many of them.  I like the idea of a mocked `fs` though. 

I also worry about page bloat.  We're loading 200kb of JS already.  It would be nice if the docs loaded instantly. (like lodash's 😏  )

megawac:
there are a bunch of things we can do to get perf wins similar to lodash site. The biggest one would obviously be implementing service workers.

Anyway, I sent a pr that should improve initial load times a bit just for you alex!

hargasinski:
If RunKit is loaded after the body like @megawac suggested, it shouldn't have an impact on the loading the page. That being said, after I finish up with this PR, I'll focus on optimizing the docs more, as right now I've mainly been looking at what features to add. I've been meaning to fix the doc generation anyways (move things over to the template from `jsdoc-fix-html.js`), so I could look into how to optimize the actual generated docs at the same time.

megawac:
fyi https://tools.pingdom.com/#!/dYSFU8/http://caolan.github.io/async/docs.html

obvious optimizations remaining include using rawgit cdn or similar for the remaining css assets and images.

aearly:
One observation:  You need to give the results area a fixed height with CSS.  Each console.log statement causes a really slow page reflow as the logging area grows.

hargasinski:
Thanks, I'll make sure to fix that. I need to add a few more tweaks to make the RunKit integration smoother before it's ready to merge.

megawac:
Codepen now supports prefilling a demo. We can also consider implementing this instead

https://blog.codepen.io/documentation/api/prefill/

aearly:
Whichever performs well when there are ~80 examples on a page.

aearly:
I made a docs site similar to this before -- one strategy I used was to only initialize the example once you scrolled to it.

PR #1320, opened 2016-11-11
Pr/1297

megawac:
/cc @CodeMan99 we decided to avoid importing lodash/filter in order to avoid including a bunch of unused deep property handling methods (that let lodash support `_.filter(arr, 'a.b.c')`) in our distribution files

Fixes #1297

### DIFF
--- a/lib/internal/filter.js
+++ b/lib/internal/filter.js
@@ -1,27 +1,29 @@
-import isArrayLike from 'lodash/isArrayLike';
+import arrayMap from 'lodash/_arrayMap';
+import property from 'lodash/_baseProperty';
 import noop from 'lodash/noop';
-
 import once from './once';
-import iterator from './iterator';
 
-export default function _filter(eachfn, coll, iteratee, callback) {
+export default function _filter(eachfn, arr, iteratee, callback) {
     callback = once(callback || noop);
-    var truthValues = isArrayLike(coll) ? new Array(coll.length) : {};
-    eachfn(coll, function (x, index, callback) {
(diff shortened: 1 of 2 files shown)


lib/internal/filter.js
```diff
@@ -1,31 +1,27 @@
...
-        }
-        else {
-            callback(null, arrayMap(results.sort(function (a, b) {
-                return a.index - b.index;
-            }), property('value')));
+        if (err) return callback(err);
+        var result = [];
+        var nextElem = iterator(coll);
+        var elem;
+        while ((elem = nextElem()) !== null) {
```
megawac: So there is a drawback/regression for doing it this way:

When `coll` is an iterator we will be trying to instantiate the iterator twice; sometimes an iterator can only be run once which would cause us to return an empty list.

megawac:
I've decided to revert the implementation change due to edge case handling (iterating iterables and objects) but will merge this for the test cases

CodeMan99:
Woo! Test cases.

megawac:
Thanks for trying to get this refactor in @CodeMan99 sorry it didn't work out. Regardless these test cases should help for the future

CodeMan99:
Yeah, it was one of those cases where I wrote a solution with with `async.map` and wondered if I could reduce the number loops with another tool. At which point I landed on `async.filter` but was pretty confused when I read the source and saw the sorting. Then I set off trying to reimplement to make it faster. You know the rest of the story from there.

At any rate, it was nice working with you. 🎩

PR #1321, opened 2016-11-12
Retry flaky tests several times when they fail

megawac:
http://mochajs.org/#retry-tests

The tests that I've added retries to are ones I've noticed fail intermittently (especially queue.pause). Let me know if any others should be added

(#1322)


aearly:
This is a good stopgap, but I'd rather refactor the tests to not need timeouts.

megawac:
Yep for sure, i created a second issue regarding the test failures

PR #1327, opened 2016-11-17
Cleanup some of the dependency imports

megawac:
Saves 800 bytes on min.js

fyi _baseRest imports a bunch of string manipulation code to manage setting `func.toString` on the output function

### DIFF
--- a/lib/autoInject.js
+++ b/lib/autoInject.js
@@ -1,7 +1,6 @@
 import auto from './auto';
 import forOwn from 'lodash/_baseForOwn';
 import arrayMap from 'lodash/_arrayMap';
-import clone from 'lodash/_copyArray';
 import isArray from 'lodash/isArray';
 import trim from 'lodash/trim';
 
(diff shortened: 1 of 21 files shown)


lib/internal/map.js
```diff
@@ -1,8 +1,7 @@
 import noop from 'lodash/noop';
-import once from './once';
 
 export default function _asyncMap(eachfn, arr, iteratee, callback) {
-    callback = once(callback || noop);
+    callback = callback || noop;
```
aearly: Don't we need `once` here?

lib/autoInject.js
```diff
@@ -109,8 +108,8 @@ export default function autoInject(tasks, callback) {
         var params;
 
         if (isArray(taskFn)) {
-            params = clone(taskFn);
-            taskFn = params.pop();
+            params = taskFn.slice(0, -1);
```
aearly: TIL `slice` can take negative indexes.

lib/internal/map.js
megawac: nope, `each`/`eachLimit`/`eachSeries` gaurentee that we'll only ever be calling callback once

PR #1334, opened 2016-11-30
Pr/1267

megawac:
The reason I prefer this approach is conceptually we're not tracking whether the callback the user provides is synchronous, rather we're just checking if we're already trying to run this function. I'll admit that this ends up doing virtually the same thing as #1267,  to me, it is just easier to reason why its necessary

/cc @autopulated @aearly for thoughts


megawac:
Specifically, I don't want to track if user functions are synchronous as I believe this will end up being a can of worms, in this case, its whether a recursive function is synchronous or async so I can support implementing this check

autopulated:
Ah this definitely looks neater 👍

Issue #1339, opened 2016-12-09
Searching aliases in documentation


megawac:
@hargasinski would it be possible to search aliases in the documentation search bar? For example, I always search for `find` and then remember that its called `detect`

hargasinski:
Definitely. I'll look into it this evening/weekend. Thanks for the heads up.

megawac:
Awesome thanks! Just bringing it up because I saw a couple SO posts with confusion on some of the aliases

hargasinski:
Quick question, did you notice the source files not appearing when searching before? For some reason, typeahead would occasionally throw an error for me about `setImmediate` not being function. I know typeahead hasn't been updated for awhile, so I don't know what would've changed. I just added a reference to `async.setImmediate` and the problem was fixed, but it still seemed really odd.

Aside from that it was a quick fix. I don't know if I should add the aliases to the toc now too (see #1340).

PR #1340, opened 2016-12-11
Add aliases to toc (related to #1339)

hargasinski:
I styled the aliases slightly differently in the toc so there's a clearer distinction. I just wanted a second opinion because it seems like a good idea, but I don't know if it adds unnecessary clutter to the toc. 
image
We already have the aliases documented in the body, and in searching.

Also, initially, it was highlighting both the alias and actual method when you clicked on an alias (due to scrollspy), which seemed weird. I changed it to just scroll the toc to the actual method, but I'm still not sure if that's the expected behaviour. Thoughts on that? The live version is on http://hargasinski.github.io/async/docs.html

Thanks in advance!

### DIFF
--- a/lib/nextTick.js
+++ b/lib/nextTick.js
@@ -14,7 +14,6 @@ import { hasNextTick, hasSetImmediate, fallback, wrap }  from './internal/setImm
  * @static
  * @memberOf module:Utils
  * @method
- * @alias setImmediate
  * @category Util
  * @param {Function} callback - The function to call on a later loop around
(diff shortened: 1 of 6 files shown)


megawac:
I don't think I like this. The toc as it is currently is quite nice as it shows how small the API of async is. I feel if that list gets too big it may intimidate new users

hargasinski:
Yeah, I think I agree. Its too busy with the aliases added. Thanks!

PR #1349, opened 2017-01-08
Print the name of the actual missing dependency in `async.auto`

jahraphael:
Hi, I'm not sure if there were specific guidelines to go through before submitting a PR, so apologies if I've missed a step. Now on to the PR:

We've been making use of `async.autoInject` as a light dependency injection framework, and whenever we'd make a mistake in a tasks dependencies we'd end up with a time-consuming task to track down what dependency `async` was failing on, because the error message wasn't giving enough to go on.

This PR simply adds the actual failed dependency to the error message to trivialize the work of tracking down missing dependencies.

### DIFF
--- a/lib/auto.js
+++ b/lib/auto.js
@@ -137,7 +137,8 @@ export default function (tasks, concurrency, callback) {
         arrayEach(dependencies, function (dependencyName) {
             if (!tasks[dependencyName]) {
                 throw new Error('async.auto task `' + key +
-                    '` has a non-existent dependency in ' +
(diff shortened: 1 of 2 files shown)


aearly:
Can you tweak the test so it checks for the dependency name as well?

jahraphael:
Sure, no problem.

I've pushed up changes that check within the `auto prevent dead-locks due to inexistant dependencies` test whether the error contains the missing dependency name (```/dependency `noexist`/)```.

megawac:
Thanks a bunch!

Issue #1348, opened 2017-01-07
Feature Request: Partition


goughy000:
Would be good to have a new collections method for "partition"

Here's a description of how the method works in Ruby:
Returns two arrays, the first containing the elements of enum for which the block evaluates to true, the second containing the rest.

So it's a bit like a filter/reject combo.

If others are interested and I get time then I'll have a go at implementing but it won't be for a few weeks at least but I thought I would put the idea here in case someone else has time

megawac:
I'm not sure how I feel about adding partition to async but I would be open to a [`groupBy`](https://lodash.com/docs/4.17.4#groupBy) which could be used to fill almost the same need. The reason I'm not a big fan of supporting partition as its slightly too specific of a use case and for methods like this we have to add `partition`, `partionLimit` and `partionSeries`

goughy000:
@megawac that makes perfect sense to me!

hargasinski:
👍 for supporting `groupBy`.

hargasinski:
Fixed in #1364.

Issue #214, opened 2013-01-11
Add "groupBy" method


katanacrimson:
Proposed method: 

Similar to sortBy, but instead of just sorting by the returned value, it will rebuild the list using the returned value as the entry's index for an array of entries.  This would be useful when trying to group or categorize.

e.g.

```
  { pid: '2384',
    name: 'node run',
    exe: 'node.exe',
    path: 'C:\\Program Files\\nodejs\\node.exe',
    cmdline: 'node.exe',
    memuse: '13348864' },
  { pid: '5000',
    name: 'C:\\Windows\\system32\\wbem\\wmiprvse.exe',
    exe: 'wmiprvse.exe',
    path: 'C:\\Windows\\system32\\wbem\\wmiprvse.exe',
    cmdline: 'WmiPrvSE.exe',
    memuse: '6266880' } ]
```

with a groupBy method, you'd be able to group individual processes by 'exe' - something that nothing else (that I can see) would allow for.

Jarwain:
Discovering this ticket 6 years later, it looks like the feature creep has been officially enshrined

PR #1364, opened 2017-02-06
[#1348] initial groupBy implementation

hargasinski:
My attempt at implementing groupBy for #1348 (and I guess #214). This is my first feature implementation for `async`, so feedback is greatly appreciated. 

I may have been a little excessive with the testing, as I ported a lot of the generic tests (like ensuring `groupByLimit` follows the concurrency limit) from `mocha_test/map.js`, while including a few specific `groupBy` tests. Also, the example I used for the docs isn't that great as I couldn't really come up with anything concise but illustrative. I'd like to change it before merging, if anyone has any ideas.

### DIFF
--- a/lib/groupBy.js
+++ b/lib/groupBy.js
@@ -1,0 +1,37 @@
+import doLimit from './internal/doLimit';
+import groupByLimit from './groupByLimit';
+
+/**
+ * Returns a new object, where each value corresponds to an array of items, from
+ * `coll`, that returned the corresponding key. That is, the keys of the object
+ * correspond to the values passed to the `iteratee` callback. Note: Since this
+ * function applies the `iteratee` to each item in parallel, there is no
+ * guarantee that the `iteratee` functions will complete in order and there is
+ * no guarantee that grouped items will appear in the same order as in `coll`.
+ *
+ * @name groupBy
+ * @static
+ * @memberOf module:Collections
+ * @method
+ * @category Collection
+ * @param {Array|Iterable|Object} coll - A collection to iterate over.
+ * @param {Function} iteratee - A function to apply to each item in `coll`.
+ * The iteratee is passed a `callback(err, key)` which must be called once it
+ * has completed with an error (which can be `null`) and the `key` to group the
+ * value under. Invoked with (value, callback).
+ * @param {Function} [callback] - A callback which is called when all `iteratee`
+ * functions have finished, or an error occurs. Result is an `Object` whoses
+ * properties are arrays of values which returned the corresponding key.
+ * @example
+ *
+ * async.groupBy(['userId1', 'userId2', 'userId3'], function(userId, callback) {
+ *     db.findById(userId, function(err, user) {
+ *         if (err) return callback(err);
+ *         return callback(null, user.age);
+ *     });
+ * }, function(err, result) {
+ *     // result is object containing the userIds grouped by age
+ *     // e.g. { 30: ['userId1', 'userId3'], 42: ['userId2']};
+ * });
+ */
+export default doLimit(groupByLimit, Infinity);
--- a/lib/groupBy.js
+++ b/lib/groupBy.js
@@ -4,10 +4,13 @@ import groupByLimit from './groupByLimit';
 /**
  * Returns a new object, where each value corresponds to an array of items, from
  * `coll`, that returned the corresponding key. That is, the keys of the object
- * correspond to the values passed to the `iteratee` callback. Note: Since this
- * function applies the `iteratee` to each item in parallel, there is no
- * guarantee that the `iteratee` functions will complete in order and there is
- * no guarantee that grouped items will appear in the same order as in `coll`.
+ * correspond to the values passed to the `iteratee` callback.
+ *
+ * Note: Since this function applies the `iteratee` to each item in parallel,
+ * there is no guarantee that the `iteratee` functions will complete in order.
+ * However, the values for each key in the `result` will be in the same order as
+ * the original `coll`. For Objects, the values will roughly be in the order of
+ * the original Objects' keys (but this can vary across JavaScript engines).
  *
  * @name groupBy
  * @static
--- a/lib/groupByLimit.js
+++ b/lib/groupByLimit.js
@@ -1,5 +1,5 @@
 import noop from 'lodash/noop';
-import doParallelLimit from './internal/doParallelLimit';
+import mapLimit from './mapLimit';
 
 /**
  * The same as [`groupBy`]{@link module:Collections.groupBy} but runs a maximum of `limit` async operations at a time.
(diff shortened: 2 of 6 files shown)


lib/groupBy.js
```diff
@@ -0,0 +1,37 @@
...
+ * @category Collection
+ * @param {Array|Iterable|Object} coll - A collection to iterate over.
+ * @param {Function} iteratee - A function to apply to each item in `coll`.
+ * The iteratee is passed a `callback(err, key)` which must be called once it
+ * has completed with an error (which can be `null`) and the `key` to group the
+ * value under. Invoked with (value, callback).
+ * @param {Function} [callback] - A callback which is called when all `iteratee`
+ * functions have finished, or an error occurs. Result is an `Object` whoses
+ * properties are arrays of values which returned the corresponding key.
+ * @example
```
aearly: I think this is a decent example.

lib/groupByLimit.js
```diff
@@ -0,0 +1,44 @@
...
+ * has completed with an error (which can be `null`) and the `key` to group the
+ * value under. Invoked with (value, callback).
+ * @param {Function} [callback] - A callback which is called when all `iteratee`
+ * functions have finished, or an error occurs. Result is an `Object` whoses
+ * properties are arrays of values which returned the corresponding key.
+ */
+export default doParallelLimit(function(eachFn, coll, iteratee, callback) {
+    callback = callback || noop;
+    coll = coll || [];
+    var result = {};
```
aearly: What do you think about having `result = Object.create(null)`?  Then instead of `hasOwnProperty.call(...)` down below, you could just do `key in result`.

mocha_test/groupBy.js
```diff
@@ -0,0 +1,375 @@
...
+                assert(false, 'iteratee should not be called');
+                next();
+            }, function(err, result) {
+                expect(err).to.eql(null);
+                expect(result).to.eql({});
+                done();
+            });
+        });
+
+        it('with reflect', function(done) {
```
aearly: I don't think the `reflect` tests are useful here.

lib/groupByLimit.js
megawac: I prefer to avoid using `Object.create(null)` as the behaviour can not be emulated consistently in older environments. I prefer using standard objects + hasOwn (hasOwn tends to be faster anyway last time I checked)

lib/groupBy.js
```diff
@@ -0,0 +1,37 @@
+import doLimit from './internal/doLimit';
+import groupByLimit from './groupByLimit';
+
+/**
+ * Returns a new object, where each value corresponds to an array of items, from
+ * `coll`, that returned the corresponding key. That is, the keys of the object
+ * correspond to the values passed to the `iteratee` callback. Note: Since this
+ * function applies the `iteratee` to each item in parallel, there is no
+ * guarantee that the `iteratee` functions will complete in order and there is
+ * no guarantee that grouped items will appear in the same order as in `coll`.
```
megawac: Reconsidering about `groupBy` again I question whether its better implemented synchronously after all the asynchronous work has been resolved, as making the order fixed in an asynchronous processing flow would make things much more complicated

mocha_test/groupBy.js
```diff
@@ -0,0 +1,375 @@
...
+    this.timeout(250);
+
+    function groupByIteratee(callOrder, val, next) {
+        setTimeout(function() {
+            callOrder.push(val);
+            next(null, val+1);
+        }, val * 25);
+    }
+
+    context('groupBy', function() {
```
megawac: Do we use context anywhere else?

lib/groupByLimit.js
hargasinski: I wouldn't mind using `Object.create(null)` internally, but I'd also prefer to avoid using it for objects returned to the user. The user might use something like `result.hasOwnProperty(anImportantKey)` somewhere later in the code, which would result in a `TypeError`.

mocha_test/groupBy.js
hargasinski: Good point, I'll remove them, thanks!

lib/groupBy.js
hargasinski: Just to make sure I understand, do you mean kinda like `./internal/filter.js` where it asynchronously transforms the values, but synchronously filters them afterwards? I could look into that tomorrow.

mocha_test/groupBy.js
hargasinski: Yes, I was looking at the [mapValues tests](https://github.com/caolan/async/blob/master/mocha_test/mapValues.js) when I added that. We use it in the `compose` and `queue` tests too.

It's more personal preference than anything, but I feel like it makes the test results easier to read and it forces them to be more organized:
image

lib/groupByLimit.js
aearly: Object.create(null) is ES5, which is our minimum required version as of 2.0.  @hargasinski brings up a good point though, best not to use it because of that.

hargasinski:
Updated the PR to use a synchronous `groupBy` implementation.

/cc @megawac @aearly

lib/groupByLimit.js
```diff
@@ -0,0 +1,51 @@
...
+            if (err) return callback(err);
+            return callback(null, {key: key, val: val});
+        });
+    }, function(err, mapResults) {
+        var result = {};
+        // from MDN, handle object having an `hasOwnProperty` prop
+        var hasOwnProperty = Object.prototype.hasOwnProperty;
+
+        for (var i = 0; i < mapResults.length; i++) {
+            if (mapResults[i]) {
```
hargasinski: Code style question: is there a preference for:
```js
if (!mapResults[i]) continue;

// do something ...
```
versus
```js
if (mapResults[i]) {
    // do something ...
}
```

aearly: I like the `continue` style, but that's just like... ...my opinion, man.

aearly:
BTW, the CI failures for this were bogus, so I'd say this is :+1: to merge.

hargasinski:
I wanted to make sure @megawac is good with the implementation before merging.

hargasinski:
@aearly @megawac should we publish `v2.2.0`?

aearly:
Yep, there's a couple other enhancements ready too.

aearly:
@hargasinski I'll let you do the honors, otherwise I'll do it later this evening.

hargasinski:
@aearly thank you, but I don't have permission to publish `async` (I'm not listed as a collaborator on npm). You'll have to do it :sweat_smile:. I did update the changelog for you for `v2.2.0` though.

aearly:
Ok, published!  Thanks for writing the changelog.

Kikobeats:
Just to know, there is no documentation for this?

aearly:
There was an error publishing docs.  Are we using the gh-pages branch, or the docs/ folder now?  Either way, `make publish-docs` needs to be fixed.

hargasinski:
I just updated the `gh-pages`, I'll look into `make publish-docs` now. Thanks for catching this @Kikobeats.

hargasinski:
@aearly rerunning `npm install; npm install gh-pages-deploy -g` appears to have resolved the issue for me. 

I couldn't figure out what was causing the exact issue (it might have been a bug in an older version of `duplexify` they were using), so I looked into alternatives just in case. The simplest one I found is [`git-directory-deploy`](https://github.com/lukekarrys/git-directory-deploy). It's similar to `gh-pages-deploy`, though it has two downsides: it doesn't confirm the commands beforehand, and it appends the given message to previous commit's message. The command would be: `git-directory-deploy --directory docs --branch gh-pages --message "gh-pages update"`. We could also use `git` directly as suggested by this [`gist`](https://gist.github.com/cobyism/4730490). The downside to that is it requires tracking the `docs` folder. 

For now, if updating `gh-pages-deploy` works, I'd suggest we stick to that, as it appears to be the best option.

aearly:
Thanks for looking in to this.  It appears there was an incompatibility with Node 6.3 streams that was making it fail: https://github.com/mafintosh/duplexify/commit/acf662f9f4bfbb2dd045bc168210ea9425375ee6 (I think I last published with Node 6.2)

Updating it fixes things.  I say we keep with `gh-pages-deploy` with this issue resolved.

Issue #687, opened 2014-12-22
Control flow to try/error if result got?


benpptung:
Hi ... I love async!! and use a lot of it, especially, I love compose. 
Thank you for the great module!! However, I cannot find a control flow to try/error the callback, is there any control flow function can do the following thing?

That said, I have `fn1`, `fn2`, `fn3`, all of them will callback an error or result, if `fn1` get the result, then the control flow end, or continue to `fn2`, then `fn3`. For example, it might look like:

```
async.try( [ 
  function(next){
     do_first_try(param, function(err, res){
         if (err) return next(err);
         next(null, res); /// the control flow end, because we have result now
     }
  }, 
  function(next){
    do_second_try(param, function(err, res){
        if (err) return next(err);
        next(null, res); /// the control flow end, because we have result now
    }
  }, 
  function(next){
    do_third_try(param, function(err, res){
        if (err) return next(err);
        next(null, res); /// the control flow end, because we have result now
    }
  },], function( err, result) {
       /// get the result whatever it comes from first try or third try, or err if no result
})
```

benpptung:
yes :) in fact, I am using `series` doing this now, `done(res)` to stop the try/error and return the result  or `done(null)`to continue the next try, but it is looking like an anti-pattern to callback.

knoid:
+1
I'll try using `series` for now. Thanks for the tip.

aearly:
After further thought -- is this any different from `retry`?  This allows you to have a unique function for every try, but is that actually useful?  This is also similar to #568

benpptung:
I think `retry` is doing the `same task` to get a successful response from task no more than specific times before returning an error. . e.g. ping a server alive or not. I think it is different from the idea of `try/error`, in which, we can try different ways to get a successful result. e.g. the typical way to get `XMLHttpRequest` in front-end.

megawac:
Yeah 👍 I'd like this.

Issue #1248, opened 2016-07-21
Handling of errors in find & some


megawac:
Upon some reflection I don't like how we handle errors in the internal `createTester`.

When an error occurs we resolve the callback resolves immediately; however, if in the case of some or detect, if the iteratee returned truthy we would have exited `createTester` successfully.

I think the correct approach would be to continue upon error and determine whether any item was truthy.

aearly:
I think exiting immediately on an error is the "correct" thing to do here.  `_.some(arr, (v) => { throw new Error(); return true })` would throw, not return true.

If the error occurred on the iteration loop that called-back `true`, would the `some` function callback true?  In that specific case, I think the error should override the `true` value.

megawac:
Sure, but in `_.some` or `Array.prototype.some`s cases its serial and can gaurentee which item it will error on. In the async case (for not *Series) we are racing to determine whether a) a error occurs b) a test passes. Seems sketch

> If the error occurred on the iteration loop that called-back true, would the some function callback true? In that specific case, I think the error should override the true value.

No that's definitely counted as an error. When checking an individual callback, error should always takes precedence over the rest of the resolve arguments.
In `detect` and `some` however, I'm not sure if an error in one warrants a error in all.

aearly:
Ah, I see, there could be some race conditions in the parallel case, or there could be conflicting results in the parallel case.  Not sure what the correct thing to do is.  Consider this example:

```js
_.some(_.range(10), function (num) {
  if (num > 5) throw new Error();
  return num === 5;
}
```

This won't throw with Lodash, but that is just an artifact of its short-circuiting.  But should it?  Half the values would throw an error if they were iterated over.  If it were iterated in reverse, it would throw every time, even though the `some` condition would be satisfied.

I'm trying to think of a real world example with Async.  Would  you want to know that one of your iteration steps produced an error, even if the `find`/`some`/`every` condition was satisfied?

megawac:
Related #591

aearly:
Hmnn, this has been open for a long time, and I'm not convinced it's still an issue now that we have `tryEach`.

PR #1365, opened 2017-02-07
Add async.tryEach

alFReD-NSH:
This adds async.tryEach which fixes what was requested in #687. I needed something with similar requirements. My use case was that I needed to get some data, which I could get it from 3 websites(with slightly different code for each) but sometimes they go down randomly.

The name might be reconsidered. We can't use `try` because it's a reserved word. I couldn't think of anything better.

You guys might want to also review the documentation, English is not my first language.

As you can see I wrote tests covering most cases, and the coverage is 100% for the file.

### DIFF
--- a/lib/tryEach.js
+++ b/lib/tryEach.js
@@ -1,0 +1,58 @@
+import noop from 'lodash/noop';
+import eachSeries from './eachSeries';
+import rest from './internal/rest';
+
+/**
+ * It runs each task in series but stops whenever any of the functions were
+ * successful. If one of the tasks were successful, the `callback` will be
+ * passed the result of the successful task. If all tasks fail, the callback
+ * will be passed the error and result (if any) of the final attempt.
+ *
+ * @name tryEach
+ * @static
+ * @memberOf module:ControlFlow
+ * @method
+ * @category Control Flow
+ * @param {Array|Iterable|Object} tasks - A collection containing functions to
+ * run, each function is passed a `callback(err, result)` it must call on
+ * completion with an error `err` (which can be `null`) and an optional `result`
+ * value.
+ * @param {Function} [callback] - An optional callback which is called when one
+ * of the tasks has succeeded, or all have failed. It receives the `err` and
+ * `result` arguments of the last attempt at completing the `task`. Invoked with
+ * (err, results).
+ * @example
+ * async.try([
+ *     function getDataFromFirstWebsite(callback) {
+ *         // Try getting the data from the first website
+ *         callback(err, data);
+ *     },
+ *     function getDataFromSecondWebsite(callback) {
+ *         // First website failed,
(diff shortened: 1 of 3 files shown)


megawac:
Could this not also be accomplished with `find`/`detect`? This looks extremely similar

alFReD-NSH:
It doesn't say in the documentation what happens when there's an error in `detect`. I believe right now it short circuits on error.
Also note that there's no data to iterate on with my use case. There's just a series of functions. That's why it's in the control flow category. It'll be slightly more code for the user to achieve this use case with `detect`.

lib/tryEach.js
```diff
@@ -0,0 +1,63 @@
...
+    var result;
+    callback = callback || noop;
+    eachSeries(tasks, function(task, callback) {
+        task(rest(function (err, args) {
+            if (args.length <= 1) {
+                args = args[0];
+            }
+            error = err;
+            result = args;
+            callback(args);
```
hargasinski: Related to the discussion in #1248. Should this be `callback(!err)`? If the task returns an error and a partial result (e.g. `[{}]`), the function should probably move onto the next task as opposed to returning.

alFReD-NSH: You're right. It should be `callback(!err)`. I'll change it today.

hargasinski:
I would be in favour of resolving #1248 before moving forward with this PR.

lib/tryEach.js
```diff
@@ -0,0 +1,63 @@
...
+ * successful. If one of the tasks were successful, the `callback` will be
+ * passed the result of the successful task. If all tasks fail, the callback
+ * will be passed the error and result (if any) of the final attempt.
+ *
+ * @name tryEach
+ * @static
+ * @memberOf module:ControlFlow
+ * @method
+ * @category Control Flow
+ * @name series
```
hargasinski: Just an FYI, some of the JSDoc tags were duplicated here.

alFReD-NSH:
@hargasinski I have fixed the issues you mentioned :)

aearly:
I think this is really cumbersome to expect people to implement with `find`/`some`, mainly because those iterate a collection of values, rather than a collection of functions.  Here's my stab at it:

```js
function nihTryEach(fns, done) {
  var lastResult;
  var lastError;
  async.some(fns, function (fn, cb) {
    fn(function(err, result) {
      lastErr = err;
      lastResult = result;
      cb(null, !err);      
    });
  }, function (err, anySucceeded) {
    if (!anySucceeded) return done(lastError);
    done(null, lastResult);
  }
}
```

It's slightly more complicated than @alFReD-NSH 's implementation.  I also think the concerns in #1248 wouldn't affect this implementation, we can deal with those separately.

I'd say this is good to merge as-is now. :+1:

aearly:
Released in v2.4.0!

PR #1368, opened 2017-02-12
[waterfall] allow optional arguments defined in tasks

futurist:
I'm encountered the problem, that allow callback passing less arguments, and `nextTask` still work as expected.

This will not break existing code, but is a feature of flexibility.

### DIFF
--- a/lib/waterfall.js
+++ b/lib/waterfall.js
@@ -80,9 +80,12 @@ export default  function(tasks, callback) {
             nextTask(args);
         }));
 
-        args.push(taskCallback);
-
         var task = tasks[taskIndex++];
+        // calc space between task arguments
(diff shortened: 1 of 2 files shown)


megawac:
Depending on function arity in Javascript can be a sketchy practice in some cases. Consider any function that wraps another

```js
function wrapper(...args) {
   return wrappedFunction(someArg, ...args);
}
```

`wrapper.length` is 0 but there might be more args than that. This would be breaking behaviour.

Instead I recommend passing `null` or `undefined` in the optional places or instead passing an object

E.g. instead of `function (a, b, c, d, callback) {...}` do `function ({a, b, c, d}, callback) { ...}`

Issue #1322, opened 2016-11-13
Fix flakey tests


megawac:
Several of the tests (especially in the queue test suite) fail quite frequently. This is due to `setTimeout` timing issues. We could sort of address this by increasing the delays but that isn't a proper solution. These tests probably just need rewrites.

These issues have been observed several times in the past

- #1152 (blocking sauce integration)
- #1303, #1311  (citgm fails intermittently) 
- frequently in travis build failures where we have to manually retrigger the build

I created PR #1321 as a stopgap, to alleviate some of the issues a bit by retrying the flakier tests when they fail. All `retries` we implement should be removed along with this fix for this issue as retries in unit tests are really ugly and may be masking other issues with the library.

Note: this may be difficult to reproduce on your local system. I've mainly noticed it in simulated environments with slower systems causing `setTimeout` calls to be delayed. Systems like travis/sauce fail quite frequently

aearly:
The [`mapValues`](https://github.com/caolan/async/blob/master/mocha_test/mapValues.js)  tests were written in such a way that they verify concurrency without using timeouts.  Someone could use a similar strategy for refactoring tests that use timeouts.

aearly:
Got some interesting new failures here: https://travis-ci.org/caolan/async/builds/220394697

Don't know what happened in the first `applyEach` test.  Also fascinating to see that `setTimeout(50) + setTimeout(50)` = `99` in the `retry` test.

aearly:
```
  asyncify
    promisified
      bluebird
        ✔ resolve
WARN: '%cWarning: a promise was rejected with a non-error: [object String]
    (No stack trace)
```

aearly:
```
  parallel
    ✖ parallel object
      Firefox 52.0.0 (Linux 0.0.0)
    Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
```

aearly:
```
FAILED TESTS:
  cargo
    ✖ without callback
      Firefox 52.0.0 (Linux 0.0.0)
    AssertionError: expected [ Array(3) ] to deeply equal [ Array(4) ] (:0)
```

hargasinski:
```
FAILED TESTS:
  cargo
    ✖ bulk task
      Firefox 52.0.0 (Linux 0.0.0)
    AssertionError: expected [ Array(4) ] to deeply equal [ Array(6) ] (:0)
```

hargasinski:
[...]

PR #1381, opened 2017-03-13
Make async.transform actually support 2 arguments

webbiesdk:
The documentation suggests that both the `accumulator` and `callback` arguments to transform is optional. 
(https://caolan.github.io/async/docs.html#transform) 

However, if both are left out (and transform therefore is called with only 2 arguments), transform crashes with a `iteratee is not a function` exception. 

By changing the equality check to a less than check, the transform function support the use-case where only 2 arguments are passed.

### DIFF
--- a/lib/transform.js
+++ b/lib/transform.js
@@ -50,7 +50,7 @@ import once from './internal/once';
  * })
  */
 export default function transform (coll, accumulator, iteratee, callback) {
-    if (arguments.length === 3) {
+    if (arguments.length <= 3) {
         callback = iteratee;
         iteratee = accumulator;
         accumulator = isArray(coll) ? [] : {};
--- a/mocha_test/filter.js
+++ b/mocha_test/filter.js
@@ -172,4 +172,17 @@ describe("reject", function () {
         });
     });
 
+    it('filter fails', function(done) {
+        async.filter({a: 1, b: 2, c: 3}, function (item, callback) {
+            if (item === 3) {
+                callback("error", false);
+            } else {
+                callback(null, true);
+            }
+        }, function (err, res) {
+            expect(err).to.equal("error");
+            expect(res).to.equal(undefined);
+            done();
+        })
+    });
 });
--- a/mocha_test/transform.js
+++ b/mocha_test/transform.js
@@ -51,4 +51,15 @@ describe('transform', function() {
             done();
         });
     });
+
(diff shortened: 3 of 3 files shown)


webbiesdk:
I have no idea why the two builds are failing. 

The coverage test claims that some internal filter.js has less coverage. Even though my change doesn't affect that. 

And somehow a test-case failed, after I added a test-case in another file.

webbiesdk:
I just triggered a rebuild by closing and re-opening the issue, and now the tests are passing. 

You guys have an unstable test!

megawac:
Yeah, its a reoccurring issue (#1322) sorry about that.

As for the PR, sure `transform` could support taking a source to iterate and an iteratee method without a final callback but what would be the point?

megawac:
Also the test failing looks to be coveralls complaining that the coverage of the repo decreased for some reason

webbiesdk:
> As for the PR, sure transform could support taking a source to iterate and an iteratee method without a final callback but what would be the point?

The point would be that your [current documentation](https://caolan.github.io/async/docs.html#transform) states that `callback` is optional. 
The current implementation actually supports an optional `callback`, but only if the number of arguments is exactly 3.

webbiesdk:
> Also the test failing looks to be coveralls complaining that the coverage of the repo decreased for some reason

The test that failed until i triggered the rebuild was: `retry with interval when some attempts fail and error test returns false at some invokation` (but it is already marked as flaky in the test-source). 

I also tried to add a new test to cover a bit of the uncovered code, and now the coverage result makes no sense, since it states that every file with changed coverage has the same or increased coverage (but somehow the overall coverage is still down). 

I also tried to run `npm run coverage" with and without my change, and it reports the exact same coverage in both cases. 

The decreased coverage is caused by a change in the environment.
To confirm that i looked at the last coverage report that gave high coverage. Just try to look at these two: 

https://coveralls.io/builds/10391324/source?filename=lib%2Finternal%2FsetImmediate.js (the last time coverage was run on master)
https://coveralls.io/builds/10579695/source?filename=lib%2Finternal%2FsetImmediate.js (coverage test after my change)

megawac:
Thanks, yeah the retry test failing is a known issue. Thanks for pointing out the docs were inaccurate.

I'll look into this coverall issue issue later, seems odd its complaining when you've added tests.

Thanks for the pull request!

Issue #942, opened 2015-10-25
"Reflect" behavior for collecting errors/results and continuing.


aearly:
Related to #349 #334 #531 .

There have been many requests for a behavior where when an iterator calls back with an error, the error is collected, and execution continues.  There have been a couple attempts at this feature, but the implementations fall short.

In theory, this would be applied to just about any async function, `map`, `series`, `parallel`, `forEachSeries`, etc..  I'd like to support this broadly, but without having to create duplicate implementations for each function.  Something we can wrap existing functions with, without having to modify the implementations.

We also need to settle on what the name is for this type of behavior.  These have been proposed:

*  `mapCollect` / `parallelCollect` / `eachSeriesCollect`
*  `mapContinue` / `parallelContinue` / `eachSeriesContinue`
*  `mapAnyway` / `parallelAnyway` / `eachSeriesAnyway`
*  `mapMaybe` / `parallelMaybe` / `eachSeriesMaybe`
*  `mapResume` / `parallelResume` / `eachSeriesResume`

Lastly -- how should the collected results  be passed to the final callback?

*  `callback(null, {errors: [], results: []])`
*  `callback(null, [Error, results, results, Error, results..])`
*  `callback(null, [{err: null, result: results0}, {err: Error, result: null}, ...])`
*  `callback(null, results, errors)`

megawac:
I would suggest `callback(null, results, errors)` or `callback(null, [Error, obj, obj, Error, etc..])`

tommy31:
Any update of this?

aearly:
Sorry, we've been too busy to make any progress on this.

jtwebman:
I am going to give this a shot as I need it for work :smile:!

jtwebman:
Actually bluebird is switching to a map with a reflect wrapper around each. Maybe that would be easier? So reflect would wrap each of the callbacks and always return successful. The output would then have null as error and a list of reflect objects that with have a function `hasValue()` method that returns true if the reflect `.value` property has a value (ie. didn't error) and if it did error then the `.error` property will have the error.

jtwebman:
We can add a reflectAll function as well that can take a list of functions with callbacks and wrap them all in async.reflect like this:

```javascript
let tasks = [
  function(callback){
    setTimeout(function(){
        callback(null, 'one');
    }, 200);
  },
  function(callback){
    // do some more stuff but error ...
    callback(new Error('bad stuff happened'));
  }
  function(callback){
    setTimeout(function(){
[...]

Issue #798, opened 2015-06-19
async.parallel - allowing all functions to finish even if error is thrown


ORESoftware:
Hello, thanks for your wonderful lib !!

One question I have - is it possible to pass async.parallel a parameter to tell it to not immediately call the final callback when there is an error passed ```callback(err,null)``` by one of the parallel functions?

I want to be able to put all the errors in an array + I also want to allow all the functions to complete, even in the case of an error in one or more of them.

Does this make sense? is it already possible?

aearly:
Very similar to #531 and  #675.  We're still deciding on the best way to do this, and what the API should look like...

ORESoftware:
all you need is optional boolean as first arg!!

:)

that's my thinking after thinking about it for 3 seconds

Issue #675, opened 2014-12-01
Looking for a way to not exit on async.parallel on error.


reggi:
Majority of my most recent use-cases involve async functions that aren't dependent on each other, and so when some async function breaks up top, the rest of my code doesn't execute. Wondering if there's a simple way to do something like this without being so obtrusive to the library / so much overhead. In this case I'd like to return false on error, so I'm building an object with every property intact, if something fails everything is else is still there, and false takes the broken async functions place.

```
var noError = function(func) {
  return function(callback) {
    func(function(err, value) {
      if (err) return callback(null, false);
      return callback(null, value);
    });
  }
}

var parallelNoError = function(functions, callback) {
  var newFunctions = {};
  for (var func in functions) {
    newFunctions[func] = noError(functions[func]);
  }
  return async.parallel(newFunctions, callback);
}
```

@aearly Would love your thoughts on this one too.

goloroden:
+1, would love to see an option to continue on errors, too (although I don't need it for `parallel`, but for `forEach`).

danielshir:
+1 for continuing on errors.
I have encountered a case where some functions running in parallel() will require cleanup code to run after completion. The current design scheme makes it hard to do so if an error occurs.

colinmegill:
+1

oillescas:
+1

skilledDeveloper:
+1

fizerkhan:
+1

Levino:
+1

jvillanti:
+1

aearly:
Closing this in favor of #942 . There's also the PR: #1012

chandlerenterprises:
+1

PR #1385, opened 2017-03-16
Add a hint to `reflect` in the documentation of `parallel`

PierreFritsch:
How to continue *parallel* processing in case of a failing task seems to be a frequently asked question. It was mentioned in several issues:

- https://github.com/caolan/async/issues/334
- https://github.com/caolan/async/issues/675
- https://github.com/caolan/async/issues/798
- https://github.com/caolan/async/issues/942

Adding a hint to the documentation might help users of the library to find out about the answer to this question.


megawac:
Thanks! Yeah very common question thanks for the documenting it!

Issue #1386, opened 2017-03-16
`async` Function Support


aearly:
One of the elephants in the room is the new `async`/`await` support that has come to Node and Chrome, and will soon hit every other major browser.  I've been thinking about what Async can do in the `async`/`await` world.

Currently, we can adapt `async` functions by wrapping them with `asyncify`.  Since an `async` function is essentially just a function that returns a Promise, that old adapter can easily convert it to a callback-style function.  However, it leads to the somewhat absurd looking:

```js
async.mapLimit(arr, 10, async.asyncify(async (val) => {
  let foo = await doSomething(val);
  //...
 return bar;
}), done);
```

However, one of the features in the spec for `async` functions is that:

```
Object.getPrototypeOf(asyncFn)[Symbol.toStringTag] === "AsyncFunction"
```

This gives a way to easily detect (native) `async` functions.  We could use this technique to automatically `asyncify` them.  The example above becomes:

```js
async.mapLimit(arr, 10, async (val) => {
  let foo = await doSomething(val);
  //...
 return bar;
}, done);
```

...which seems to flow much more naturally.  I also think we should continue to use callbacks.  If a user wanted to `await` the result, they would have to `promisify` the function, or `pify` Async as a whole:

```js
let result = await pify(async.mapLimit)(arr, 10, async (val) => {
  let foo = await doSomething(val);
  //...
 return bar;
});
```

The above method for detecting `async` functions only works with native functions.  I don't think there is a way to detect Babel transpiled functions.  We certainly can't detect normal functions that simply return Promises, because we'd have to retroactively not pass a callback.  There would he a huge caveat that this would only work without a transpiler in very modern environments, otherwise you still have to manually wrap with `asyncify`.

Also, admittedly, many Async methods don't make sense with `async`/`await`.  Most of the control flow methods (save for things like `auto` and `queue`) are more easily replicated with native control flow constructs.  `map` and `parallel` can be replaced with `Promise.map` and `Promise.all`.  However, the limiting collection functions would be very useful, as well as `auto` and a few others.  (Also, `autoInject` with `async` functions is a async control flow dream!)

megawac:
I'm going to ponder over this some more but I'd like to ask a couple technical questions just so I can try to better envision what this looks like..

Is there a reason to do `Object.getPrototypeOf(asyncFn)[Symbol.toStringTag] === "AsyncFunction"` or can we do `asyncFn[Symbol.toStringTag] === "AsyncFunction"` (seems to work in FF)?

[...]

PR #1390, opened 2017-03-25
`async` function support

aearly:
#1386 

This adds support for passing `async` functions wherever it might make sense.

Obviously we are duplicating some core functionality with things like `map` and `race`, but this allows us to treat things uniformly.  We do have some good value to add with functions that limit concurrency, and adding more higher-order operations that just `map` and `each` (as you would get with `Promise.map()` and `Promise.all()`).  "Lodash for `async` functions" is a good thing to strive to be.

Most of the control flow functions are of dubious value, (`waterfall`, `parallel`, `whilst`), but you can do some really neat things with `auto` and `queue`.  (I really love the `autoInject` shorthand test!)

The utility functions are a bit strange.  `retry`, `reflect`, `memoize` and `timeout` now take in an `async` function and return a callback-accepting function.  It feels like those should create a Promise-returning function.


aearly:
Also, I ran the benchmarks -- there is no difference in performance, whether the environment supports async functions or not.

.travis.yml
```diff
@@ -4,27 +4,29 @@ node_js:
...
-    - node_js: "6"
-      addons:
-        firefox: "49.0"
-      env: BROWSER=true MAKE_TEST=true
+#matrix:
+#  include:
+#    - node_js: "6"
+#      addons:
+#        firefox: "49.0"
+#      env: BROWSER=true MAKE_TEST=true
```
megawac: why are we disabling this stuff? Can we just up the firefox version?

lib/autoInject.js
```diff
@@ -106,22 +108,25 @@ export default function autoInject(tasks, callback) {
...
+        var fnIsAsync = isAsync(taskFn);
 
         if (isArray(taskFn)) {
             params = taskFn.slice(0, -1);
             taskFn = taskFn[taskFn.length - 1];
 
             newTasks[key] = params.concat(params.length > 0 ? newTask : taskFn);
-        } else if (taskFn.length === 1) {
+        } else if ((!fnIsAsync && taskFn.length === 1) ||
+            (fnIsAsync && taskFn.length === 0)) {
```
megawac: `fnIsAsync + taskFn.length === 1` =P (dont do it lol)

lib/doWhilst.js
```diff
@@ -19,7 +20,7 @@ import onlyOnce from './internal/onlyOnce';
  * passes. The function is passed a `callback(err)`, which must be called once
  * it has completed with an optional `err` argument. Invoked with (callback).
  * @param {Function} test - synchronous truth test to perform after each
```
megawac: Is there a `AsyncFunction` jsdoc tag?

lib/internal/applyEach.js
```diff
@@ -1,11 +1,13 @@
...
 import rest from './rest';
 import initialParams from './initialParams';
+import wrapAsync from './wrapAsync';
 
 export default function applyEach(eachfn) {
     return rest(function(fns, args) {
         var go = initialParams(function(args, callback) {
             var that = this;
-            return eachfn(fns, function (fn, cb) {
+            return eachfn(arrayMap(fns, wrapAsync), function (fn, cb) {
```
megawac: `fns` can be a `array|object|iterable`

lib/internal/applyEach.js
```diff
@@ -1,11 +1,13 @@
...
                 fn.apply(that, args.concat(cb));
```
megawac: Do `wrapAsync` here instead

lib/autoInject.js
aearly: 🤢

lib/doWhilst.js
aearly: I was thinking about the best way to document this.  For sure, add something to the `intro.md`, but we have lots of repeated docs where we describe the iteratee function as a node-style async function.  It would be nice to centralize the description of what Async accepts, with each Async method just describing the non-callback parameters.

lib/internal/applyEach.js
aearly: Good call. 👍

.travis.yml
aearly: I disabled this thinking it was an issue with karma, but it's a syntax error.  I'll investigate further and re-enable.  We could test both 45 (ESR, no `async` support) and 52 (`async` supported).

lib/autoInject.js
aearly: Actually, I should probably factor this condition out into a `hasNoDeps` variable.

lib/internal/wrapAsync.js
```diff
@@ -0,0 +1,27 @@
+import identity from 'lodash/identity';
+import asyncify from '../asyncify';
+
+var supportsSymbol = typeof Symbol === 'function';
+
+function supportsAsync() {
+    var supported;
+    try {
+        /* eslint no-eval: 0 */
+        supported = isAsync(eval('(async function () {})'));
```
megawac: `eval` can be abused in the same way that `new Function` can be abused. I'd rather just use `eval` here

lib/doWhilst.js
megawac: Perhaps an `external` definition if jsdoc doesn't support some way of documenting an async function?

http://usejsdoc.org/tags-external.html

lib/internal/wrapAsync.js
aearly: `new Function()` and `eval` are equivalently evil.  Here, they are necessary, because there is no other way to detect `async` support without causing a parse error that halts JS execution.

lib/doWhilst.js
aearly: I'm also thinking of using a `@typedef`, like we do for the return values of `queue`/`cargo`. 

http://usejsdoc.org/tags-typedef.html

megawac: I really don't like how the `@typedef`s look in our code. I'd rather just link to MDN

aearly: Check the latest commit.  It's cleaner than the JSDoc docs make it out to be...

lib/asyncify.js
```diff
@@ -48,7 +48,7 @@ import initialParams from './internal/initialParams';
  *     }
  * ], callback);
  *
- * // es6 example
+ * // es2017 example
```
hargasinski: Would it be useful to add a note here that `async` functions are supported out of the box (when not transpiled)?

lib/doWhilst.js
hargasinski: :+1: for using `typedef` for this. Does this mean most of the docs should to be updated to reflect that `iteratee`  should be an `AsyncFunction` type?

lib/internal/wrapAsync.js
hargasinski: I'm good with using `eval` here instead of `new Function`.

lib/doWhilst.js
aearly: Yes, lots of docs to update. :sweat:   It does create a nice link to the central `AsyncFunction` docs in the rendered docs, though.

lib/asyncify.js
aearly: Yeah, we should clarify here.

lib/index.js
```diff
@@ -1,11 +1,45 @@
...
+ * Error as its first argument to signal that an error occurred.
+ * Otherwise, it can be called with `null` as the first argument, and any
+ * additional `result` arguments that may apply to signal successful completion.
+ * The callback must be called exactly once, ideally on a later tick of the
+ * JavaScript event loop.
+ *
+ * This type of function is also referred to as a "Node-style async function".
+ *
+ * Wherever we accept a Node-style async function, we also directly accept an
+ * ES2017 `async` function.
```
megawac: Link MDN here (https://developer.mozilla.org/docs/Web/JavaScript/Reference/Statements/async_function)

lib/index.js
```diff
@@ -1,11 +1,45 @@
...
+ * In this case, the `async` function will not be passed a callback, and any
+ * thrown error will be used as the `err` argument of a theoretical callback,
+ * and the return value will be used as the `result` value.
```
megawac: What do you mean `the `async` function will not be passed a callback`? Do you mean the result of the async function?

lib/index.js
```diff
@@ -1,11 +1,45 @@
...
+ *
+ * Note that we can only detect native `async function`s in this case.
```
megawac: > Note, due to JavaScript limitations, we can only detect native `async function`s and not polyfilled implementations.

aearly:
@megawac @hargasinski Many doc updates for your :eyes:.  :sweat:

.eslintrc
```diff
@@ -6,7 +6,7 @@
     "es6": true
   },
   "parserOptions": {
-    "ecmaVersion": 6,
+    "ecmaVersion": 8,
```
megawac: what is es8? can we set this to 2017/2018 whatever?

package.json
```diff
@@ -22,12 +22,13 @@
...
   "devDependencies": {
-    "babel-cli": "^6.16.0",
-    "babel-core": "^6.3.26",
+    "babel-cli": "^6.24.0",
+    "babel-core": "^6.24.0",
     "babel-plugin-add-module-exports": "^0.2.1",
     "babel-plugin-istanbul": "^2.0.1",
     "babel-plugin-transform-es2015-modules-commonjs": "^6.3.16",
     "babel-preset-es2015": "^6.3.13",
+    "babel-preset-es2017": "^6.22.0",
```
megawac: are we using this?

.eslintrc
aearly: ES8 is eslint's shorthand for ES2017.  No, you can't use "2017"...  Just subtract 2009. 😛

package.json
aearly: Yes, otherwise babel can't parse the test file where we use `async` functions, even if they're being used natively.

megawac:
👏 awesome work, pretty excited for this

hargasinski:
I'm a little late, but really great work. I'm excited about this :100:

PR #1397, opened 2017-04-07
queue.remove()

### LINKED ISSUE #1391: queue.remove() with filter function
I'm currently finding a way to remove certain unprocessed entries in a queue/priorityQueue. I don't know how to do it, so I would appreciate if anyone could give me a hint.

**Feature suggestion**: add another method with which entries gets removed like that:

    var q = async.queue(function (task, callback) {
        callback();
    });
    q.push({some: 'thing'});
    q.remove(function (task) {
        return task.some === 'thing'; // remove task if return value is true
    });

Or with the priority queue: 

    var q = async.priorityQueue(function (task, callback) {
        callback();
    }, 3);
    q.push({some: 'thing'}, 100);
    q.remove(function (task, priority) {
        return task.some === 'thing'; // remove task if return value is true
        or
        return priority === 100; // remove task with certain priority
    });

aearly:
Closes #1391 .

Adds a `remove` function to queues.  Without it, there is no way to arbitrarily remove items from the queue.  In Async 1.x and below, the `tasks` list was just an array, so you could splice as you needed.  This adds a `filter` function to the `DoublyLinkedList` class that is used by `q.remove`.

I also added some unit tests for `DLL` itself.  I also noticed that the prior implementation of `empty` could cause a memory leak.  I think the lingering chain of `prev`/`next` references could fool GC and keep the objects around.

### DIFF
--- a/lib/internal/DoublyLinkedList.js
+++ b/lib/internal/DoublyLinkedList.js
@@ -23,7 +23,10 @@ DLL.prototype.removeLink = function(node) {
     return node;
 }
 
-DLL.prototype.empty = DLL;
+DLL.prototype.empty = function () {
+    while(this.head) this.shift();
+    return this;
+};
 
 DLL.prototype.insertAfter = function(node, newNode) {
     newNode.prev = node;
@@ -60,3 +63,26 @@ DLL.prototype.shift = function() {
 DLL.prototype.pop = function() {
     return this.tail && this.removeLink(this.tail);
 };
+
+DLL.prototype.toArray = function () {
+    var arr = Array(this.length);
+    var idx = 0;
+    var curr = this.head;
+    for(idx = 0; idx < this.length; idx++) {
+        arr[idx] = curr.data;
+        curr = curr.next;
+    }
+    return arr;
+}
+
+DLL.prototype.filter = function (testFn) {
+    var curr = this.head;
+    while(!!curr) {
+        var next = curr.next;
+        if (!testFn(curr)) {
+            this.removeLink(curr);
+        }
+        curr = next;
+    }
+    return this;
+}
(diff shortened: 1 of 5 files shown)


lib/internal/DoublyLinkedList.js
```diff
@@ -60,3 +63,26 @@ DLL.prototype.shift = function() {
...
+    var idx = 0;
+    var curr = this.head;
+    for(idx = 0; idx < this.length; idx++) {
+        arr[idx] = curr.data;
+        curr = curr.next;
+    }
+    return arr;
+}
+
+DLL.prototype.filter = function (testFn) {
```
megawac: Is this destructive? If so I'd rather not call it `filter`, but `remove`

lib/internal/queue.js
```diff
@@ -102,6 +102,11 @@ export default function queue(worker, concurrency, payload) {
         unshift: function (data, callback) {
             _insert(data, true, callback);
         },
+        remove: function (testFn) {
+            q._tasks = q._tasks.filter(function (node) {
+                return !testFn(node);
+            })
```
megawac: nit: missing semi

lib/internal/DoublyLinkedList.js
aearly: I wanted to call it `filter` so that if someone wants to swap out the implementation of `_tasks` with an array, they could.  e.g.:

```js
var q = async.queue(worker);
q._tasks = [];
```

I remember we got an issue a while back after 2.0 because some people were doing custom splicing in the queue, which they couldn't do with the linked list.

It is strange than it's doesn't return a new list, though.

aearly: Actually after more thought, `remove` is a better name.  If someone is using an array implementation, they're likely doing their own array splicing somehow anyway.

PR #1398, opened 2017-04-08
Added ability to pass initial params to asyncify

Telokis:
Modifies the behavior of `async/asyncify` to match `async/apply`: we can now pass additional *initial* arguments to `asyncify` and those will be prepended to the *real* function call.

### DIFF
--- a/lib/asyncify.js
+++ b/lib/asyncify.js
@@ -1,5 +1,4 @@
-import isObject from 'lodash/isObject';
-import initialParams from './internal/initialParams';
+import isObject from 'lodash/isObject'
 
 /**
  * Take a sync function and make it async, passing its return value to a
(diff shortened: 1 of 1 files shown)


aearly:
Sorry, but I'd prefer to leave argument binding to separate functions like `Funcion.bind`, `_.partial`, or `async.apply`.  

```js
var asyncWrapped = async.asyncify(function foo(arg2, arg3) {
  //...
}).bind(null, arg0, arg1);
```

Telokis:
This is extremely verbose.
To me, `asyncify` is *exactly* `apply` but for sync functions so, when using it, I expect it to have the same capabilities.

Telokis:
@aearly This is extremely verbose.
To me, asyncify is exactly apply but for sync functions so, when using it, I expect it to have the same capabilities.

PR #1399, opened 2017-04-09
Hopefully fix flakey tests

aearly:
Related: #1322 

I completely re-did that notorious `queue.pause` test.  I also tweaked the flakey retry test, so it might give more clues in the future when it fails.


megawac:
Lets give it a try. If we dont have errors for a bit we can close #1322. It might also be worth testing it with one of the browser ci services and see if it fixes it there. The issues seemed to occur more often in the browser ci world.

aearly:
Cool.  I want to re-run the tests during the middle of a weekday when Travis is under high load.  That seems to be when tests fail more often.  I haven't gotten these tests to fail yet after a dozen or so re-runs.

PR #1405, opened 2017-04-17
Build Refactor

aearly:
This is a major refactor/simplification of the build process. There are no changes to the output build.  There are subtle whitespace/comment changes in `dist/async.js` but the minified version is identical.

Summary of changes:

* Don't use babel in the build scripts, just use Node 6
* Use the `build-es/*.js` files to build `build/dist/async.js`
* Remove the unused `build/index.js`  (`build/dist/async.js` is used instead)
* Only use babel to compile `lib/*.js` to `build/*.js`  (ES to CJS)
* Rely on Make more to manage build artifacts.  There is now no need to `make clean` before each build
* Since there is now only one use of Babel and one use of Rollup, the build scripts are much simpler

I'd recommend reviewing individual commits to see the process, this is a lot. (Although I did muddy things a bit with a rebase I shouldn't have done)


aearly:
@megawac Would love your 👀 on this so I can merge it before v2.4.0

megawac:
Yeah, I want to spend some time playing with this change before I approve. I looked over the changes and it looks good

megawac:
Not a significant issue, but I notice that `make build` is significantly slower now. Before it would make the files in parallel, now each file is built seperately

aearly:
You can specify parallelization in Make with the `-j` flag to speed it up.  e.g. `make build -j 8` to fire up all your cores.  I think there's a way to have a default value for make flags.

PR #1408, opened 2017-04-18
[asyncify] Make errors in callbacks throw globally

davidaurelio:
If asyncify wraps a function returning promises, the callback will be executed as part of the promise’s `.then()` method.
This means that any error thrown from that method will be silenced, and lead to a “unhandled promise rejection” warning in modern engines.

Usually, we want these errors to be visible, and even crash the process.

I’d like to add a test, but there doesn’t seem to be a good way to do it, especially for both node.js and browsers.

### DIFF
--- a/lib/asyncify.js
+++ b/lib/asyncify.js
@@ -1,5 +1,6 @@
 import isObject from 'lodash/isObject';
 import initialParams from './internal/initialParams';
+import setImmediate from './internal/setImmediate';
 
 /**
  * Take a sync function and make it async, passing its return value to a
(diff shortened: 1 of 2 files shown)


davidaurelio:
I’m going to fix the broken tests, but would like feedback on the change first.

aearly:
I've thought about deferring the callback in `asyncify`, for the reason you mention, but I don't like it because of the extra deferral.

Another strategy would be to try/catch the callback, and then re-throw the error on the next tick.  This is what Bluebird does:  https://github.com/petkaantonov/bluebird/blob/master/src/nodeify.js#L38-L42

Either way, if it involves changing our tests, it's a breaking change, albeit a small one.

Also, in the next version of Node, unhandled promise rejections are going to start killing the process, just like an unhandled error.

aearly:
Re. catching an async error, you can do this:  https://github.com/caolan/async/blob/master/mocha_test/waterfall.js#L110-L119

davidaurelio:
Thanks for the hints. I updated the PR.

> Another strategy would be to try/catch the callback, and then re-throw the error on the next tick

Seems the best way to do it. I adapted the code accordingly.

> Either way, if it involves changing our tests, it's a breaking change, albeit a small one.

It *is* a breaking change, but you could argue it is a fix. The behavior is now the same as when asyncify’ing functions that don’t return promises. There’s even [a test for errors to be thrown in the non-promise case](https://github.com/caolan/async/blob/master/mocha_test/asyncify.js#L49).

aearly:
You're right, after further thought, this is more of a bug fix.  It will change how people's code behaves, though.

aearly:
Released in v2.4.0!

davidaurelio:
great, thanks for letting me know!

PR #1419, opened 2017-05-21
Allow functions wrapped in `timeout` to be called multiple times (fixes #1418)

### LINKED ISSUE #1418: Result of async.timeout() only works the first time you call it.
**What version of async are you using?**  2.4.0

**Which environment did the issue occur in (Node version/browser version)**  Node.js 4.7.0

**What did you do? Please include a minimal reproducable case illustrating issue.**  
```

var asyncFn = function (i, done) {
    if (i < 1) {
        setTimeout(done, 200);
    } else {
        setImmediate(done);
    }
};

// wrap function once, call 10 times
var timeoutFn = async.timeout(asyncFn, 100);
var j = 0;
async.whilst(
    function () {
        return j < 10;
    },
    function (done) {
        timeoutFn(j++, function (err) {
            console.log('Test 1 #' + j, err);
            done();
        });
    },
    function (err) {
    }
);
```

**What did you expect to happen?**  Should only get timeout error the first time it's called.

**What was the actual result?**  Timeout occurs every time it's called, even though the underlying function only takes too long the first time.

hargasinski:
Previously, the scoping of the `timedOut` variable in `timeout` meant that if any call to wrapped function timed out, any subsequent calls would also give a `ETIMEDOUT` error. This PR fixes that by moving the variables down one scope, so they are reinitialized each call. As a side benefit, it also lets the wrapped function be passed to functions such as `async.forEach`.

### DIFF
--- a/lib/timeout.js
+++ b/lib/timeout.js
@@ -43,33 +43,32 @@ import wrapAsync from './internal/wrapAsync';
  * });
  */
 export default function timeout(asyncFn, milliseconds, info) {
-    var originalCallback, timer;
-    var timedOut = false;
-
-    function injectedCallback() {
-        if (!timedOut) {
-            originalCallback.apply(null, arguments);
-            clearTimeout(timer);
-        }
-    }
-
-    function timeoutCallback() {
-        var name = asyncFn.name || 'anonymous';
-        var error  = new Error('Callback function "' + name + '" timed out.');
-        error.code = 'ETIMEDOUT';
-        if (info) {
-            error.info = info;
-        }
-        timedOut = true;
-        originalCallback(error);
-    }
-
     var fn = wrapAsync(asyncFn);
 
-    return initialParams(function (args, origCallback) {
-        originalCallback = origCallback;
+    return initialParams(function (args, callback) {
+        var timedOut = false;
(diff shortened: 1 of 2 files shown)


lib/timeout.js
```diff
@@ -43,33 +43,32 @@ import wrapAsync from './internal/wrapAsync';
...
+            }
+            timedOut = true;
+            callback(error);
         }
-        timedOut = true;
-        originalCallback(error);
-    }
 
-    var fn = wrapAsync(asyncFn);
+        args.push(function () {
```
hargasinski: I noticed this style in [`ensureAsync`](https://github.com/caolan/async/blob/master/lib/ensureAsync.js#L44). I think I prefer it over the `args.concat(namedFunction)` style. Should I change it back to the way it was before?

aearly: This is fine.  `args.concat(namedFunction)` is nice because it's a singe expression that returns the resulti, but it also creates a new array.  `push` is fine here because nothing else re-uses the `args` array.

aearly:
Amazing that this wasn't caught until now.  It makes me feel like no one was actually using `timeout`....

PR #1420, opened 2017-05-21
may be it's better to use the length method

JoeZheng2015:
According to code in queue.js
``` javascript
length: function () {
     return q._tasks.length;
}
```
It may be better to use `q.length()` rather than `q._tasks.length`


hargasinski:
Thanks @JoeZheng2015!

Issue #1314, opened 2016-10-27
mapLimit with n/sec limit


mvalletta:
Is possible to limit parallel execution to a number/sec?
This is necessary for those services, like Google Sheet API Limits and Quotas  (see https://developers.google.com/analytics/devguides/config/mgmt/v3/limits-quotas) , that limit the number of queries per second (QPS) per IP.
Thanks!

aearly:
Similar features have been asked about before.  ( #1113 #1082 #1020 #942 #37 )  There is demand for some sort of "throttled queue" functionality, but managing that kind of construct properly is a hard problem.

megawac:
tl;dr you/async would need some sort of smoothing function to do this. If we were to do this we would probably implement [an exponential smoothing function](https://en.wikipedia.org/wiki/Exponential_smoothing)

JohnLelii-CMCST:
OP: maybe you've already solved this.....I've used another NPM package to implement rate limiting using leakybucket or token bucket models.  https://www.npmjs.com/package/limiter  This has worked well for me in the past.  To create a rate limited mapper function, just do something like this.  Your mapper function is directly used with async.

`var RateLimiter = require("limiter").RateLimiter;
var limiter = new RateLimiter(2, "second");
var async = require('async');

function rateLimitedMapper(item, callback)
{
  limiter.removeTokens(1, function(err, remainingRequests) {
    console.log("working on item: " + item);
    callback(null, item.length);
  });
}

async.map(input, rateLimitedMapper, function(err, results) {
  // do something with results
 });
`

aearly:
That seems like a lib we should mention in our readme.  Seems to do exactly what we wanted to build.

megawac:
Yep, it's using a token bucket under the hood which is the same approach I was taking. I'd be quite fine deferring rate limitting to external libraries :)

aearly:
Yeah, honestly I'm not thrilled with having to support something like that within Async.  There are so many gotchas when doing rate limiting.

aearly:
Changing the tags on this to signify that we need a place in the docs for other libraries that extend Async.

manoharreddyporeddy:
Another way to do this using Promise.all and promise-throttle:

https://stackoverflow.com/questions/20253425/throttle-and-queue-up-api-requests-due-to-per-second-cap/47627820#47627820

PR #1425, opened 2017-06-03
added rateLimit parameter to the queue

glynnbird:
I'm a heavy user of `async.queue`, but if I am using it to queue calls to an API service which implements rate limiting (say 5 API calls per second), it becomes very tricky to get the queue to consume the queue within the rate limits.

To fix this, I branched the code and added a new parameter to the queue object:

```js
var worker = function(data, done) {
  setTimeout(done, 100);
};
var concurrency = 1;
var rateLimit = 5;
var q = async.queue(worker, concurrency, rateLimit);
```

The queue works as normal but only proceeds at a maximum rate of 5 items per second. 

The existing tests all pass and supplying a rateLimit of `null` or leaving it as undefined keeps the original queue unchanged. 

I've added a test for the rate-limited version.

(as this is my first commit to this project, I'm not sure if I should be including the "dist" files in the PR - please let me know if you would like these included too).

### DIFF
--- a/lib/cargo.js
+++ b/lib/cargo.js
@@ -78,5 +78,5 @@ import queue from './internal/queue';
  * });
  */
 export default function cargo(worker, payload) {
-    return queue(worker, 1, payload);
+    return queue(worker, 1, null, payload);
 }
--- a/lib/internal/queue.js
+++ b/lib/internal/queue.js
@@ -7,7 +7,7 @@ import setImmediate from './setImmediate';
 import DLL from './DoublyLinkedList';
 import wrapAsync from './wrapAsync';
 
-export default function queue(worker, concurrency, payload) {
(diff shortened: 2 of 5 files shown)


aearly:
Sorry, but we've discussed implementing rate limiting in Async [many times](https://github.com/caolan/async/issues?q=rate+limit+is%3Aclosed) ( #1314 ), and have decided against it due to the incidental complexity/gotchas involved.

You're welcome to publish a standalone rate limiting queue if you wish.  We'll link it in our docs.

glynnbird:
Thanks for taking the time to reply @aearly. I understand perfectly why you wouldn't want to creeping the scope any further out any more ;)

I've published my code as a standalone as suggested here: https://www.npmjs.com/package/qrate

PR #1429, opened 2017-06-07
ensure q.workersList() contains items being processed [fixes #1428]

### LINKED ISSUE #1428: Size discrepancy between queue.workersList().length and queue.running()
2.4.1
Node v4.6.0
queue.workersList().length should be the same size as queue.running()
They sometimes are not the same length.

I'm observing a discrepancy between queue.workersList().length and queue.running(). It looks like lib/internal/queue.js line 60 may be the culprit. Shouldn't we always be splicing exactly one element here instead of potentially many elements. So replacing:

`workersList.splice(index)`

with 

`workersList.splice(index, 1)`

hargasinski:
Aligns `q.workersList()` and `cargo.workersList()` with the [docs](https://caolan.github.io/async/docs.html#QueueObject).

`workersList()` is not really relevant in the context of `async.cargo`, as there will only be 1 worker running at a given time, but I would prefer to keep the behavior consistent between it and `queue`.

### DIFF
--- a/mocha_test/queue.js
+++ b/mocha_test/queue.js
@@ -759,16 +759,19 @@ describe('queue', function(){
                 expect(
                     getWorkersListData(q)
                 ).to.eql(itemsBeingProcessed[task]);
+                expect(q.workersList().length).to.equal(q.running());
                 async.setImmediate(function() {
                     expect(
                         getWorkersListData(q)
                     ).to.eql(itemsBeingProcessed[task+'_cb']);
+                    expect(q.workersList().length).to.equal(q.running());
                     cb();
                 });
             }, 2);
 
             q.drain = function() {
                 expect(q.workersList()).to.eql([]);
+                expect(q.workersList().length).to.equal(q.running());
                 done();
             };
 
--- a/mocha_test/queue.js
+++ b/mocha_test/queue.js
@@ -656,7 +656,7 @@ describe('queue', function(){
         });
     });
 
-    context('q.unsaturated(): ',function() {
+    context('q.unsaturated(): ', function() {
         it('should have a default buffer property that equals 25% of the concurrenct rate', function(done){
             var calls = [];
             var q = async.queue(function(task, cb) {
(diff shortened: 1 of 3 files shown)


hargasinski:
Also, just in case, I ran the benchmarks for `queue`. There doesn't appear to be a difference.
```bash
$ node perf/benchmark.js --grep queue
Latest tag is  v2.4.1
Comparing v2.4.1 with current on Node v7.4.0
queue(1000) v2.4.1 x 581 ops/sec ±1.52% (31 runs sampled), 1.72ms per run
queue(1000) current x 606 ops/sec ±0.69% (32 runs sampled), 1.65ms per run
current is faster
queue(30000) v2.4.1 x 16.96 ops/sec ±2.82% (30 runs sampled), 59.0ms per run
queue(30000) current x 15.94 ops/sec ±2.28% (29 runs sampled), 62.7ms per run
v2.4.1 is faster
queue(100000) v2.4.1 x 4.11 ops/sec ±10.95% (9 runs sampled), 243ms per run
queue(100000) current x 4.22 ops/sec ±6.17% (9 runs sampled), 237ms per run
Tie
queue(200000) v2.4.1 x 2.10 ops/sec ±13.67% (5 runs sampled), 477ms per run
queue(200000) current x 2.07 ops/sec ±11.86% (5 runs sampled), 482ms per run
Tie
Both versions are about equal (781ms total vs. 784ms total)
Both versions won the same number of benchmarks (1 vs. 1)
```

mocha_test/queue.js
```diff
@@ -719,6 +719,65 @@ describe('queue', function(){
...
+            };
+
+            function getWorkersListData(q) {
+                return q.workersList().map(function(v) {
+                    return v.data;
+                });
+            }
+
+            var q = async.queue(function(task, cb) {
+                expect(
```
aearly: Should we also assert that `workersList().length === running()` here?

hargasinski: Added, thanks!

PR #1430, opened 2017-06-07
add concatLimit [fixes #1426]

### LINKED ISSUE #1426: Elusive `concatLimit`
Some issues and PRs (#973 #847) mention `concatLimit` but it's nowhere to be found in actual API and docs. Would be neat to have it, or is there a reason it is not provided? (I haven't been able to find anyting in changelog or otherwise myself.)

Thanks!

hargasinski:
Adds the elusive `concatLimit`.

### DIFF
--- a/lib/concatLimit.js
+++ b/lib/concatLimit.js
@@ -1,0 +1,22 @@
+import doParallelLimit from './internal/doParallelLimit';
+import concat from './internal/concat';
+
+/**
+ * The same as [`concat`]{@link module:Collections.concat} but runs a maximum of `limit` async operations at a time.
+ *
+ * @name concatLimit
+ * @static
+ * @memberOf module:Collections
+ * @method
+ * @see [async.concat]{@link module:Collections.concat}
+ * @category Collection
+ * @param {Array|Iterable|Object} coll - A collection to iterate over.
+ * @param {number} limit - The maximum number of async operations at a time.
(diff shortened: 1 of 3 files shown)


hargasinski:
While I was playing around with the `concat` function, I noticed it doesn't preserve order. i.e.
```js
async.concat([30, 15], function(val, cb) {
    setTimeout(function() {
        cb(null, [''+val]);
    }, val);
}, function(err, result) {
    console.log(result);
});
```
will log
```js
[ '15', '30' ]
```
as opposed to
```js
[ '30', '15' ]
```

Should this be changed to the match the behavior of functions like [filter](https://caolan.github.io/async/docs.html#filter), [groupBy](https://caolan.github.io/async/docs.html#groupBy) or [map](https://caolan.github.io/async/docs.html#map), which preserve the order in `result`? Would that be considered a breaking change?

/cc @aearly @megawac

aearly:
You're right -- it should be order-preserving.  I think that would be a subtle breaking change, but perhaps not worth saving for 3.0. I think `concat` has very few users (judging by how long it took someone to notice `concatLimit` missing).

hargasinski:
@aearly after thinking about it a little more, it could technically fall under a bug fix, i.e. a patch release. Either way, I'll open a PR for it either tonight or at some point tomorrow. Thanks!

riska8899 (approved):

PR #1436, opened 2017-06-12
preserve order, make variadic and handle falsy values in concat

hargasinski:
This is based on the discussion in #1430. It adds order preservation to the `async.concat` functions.

<s>I feel like I might have changed our internal structure a little bit too much though by removing `internal/concat` and `internal/doSeries`. If I did, let me know, I'll change the PR to only modify the tests and `internal/concat`. Thanks!</s> This isn't actually changing our public API, so I feel like it should be okay?

<s>Otherwise, I still want to add a couple more tests before merging (for sparse results, original untouched, etc...).</s> Added tests.


hargasinski:
As a side note, the performance decrease is pretty significant:
```bash
$ node perf/benchmark.js --grep concat
Latest tag is  v2.4.1
Comparing v2.4.1 with current on Node v7.10.0
concat(10) v2.4.1 x 74,243 ops/sec ±1.04% (28 runs sampled), 0.0135ms per run
concat(10) current x 52,714 ops/sec ±1.35% (31 runs sampled), 0.0190ms per run
v2.4.1 is faster
concat(300) v2.4.1 x 4,029 ops/sec ±2.68% (30 runs sampled), 0.248ms per run
concat(300) current x 2,748 ops/sec ±1.29% (32 runs sampled), 0.364ms per run
v2.4.1 is faster
concat(10000) v2.4.1 x 9.86 ops/sec ±2.20% (19 runs sampled), 101ms per run
concat(10000) current x 8.69 ops/sec ±3.27% (17 runs sampled), 115ms per run
v2.4.1 is faster
v2.4.1 faster overall (102ms total vs. 115ms total)
v2.4.1 won more benchmarks (3 vs. 0)
```

lib/concatSeries.js
```diff
@@ -19,4 +19,4 @@ import doSeries from './internal/doSeries';
  * containing the concatenated results of the `iteratee` function. Invoked with
  * (err, results).
  */
-export default doSeries(concat);
```
megawac: is this the last use of `doSeries`?

mocha_test/groupBy.js
```diff
@@ -330,7 +330,7 @@ describe('groupBy', function() {
         });
 
         it('handles empty object', function(done) {
-            async.groupByLimit({}, 2, function(val, next) {
+            async.groupBySeries({}, function(val, next) {
```
megawac: why this test change?

mocha_test/slice.js
```diff
@@ -0,0 +1,32 @@
+var slice = require('../lib/internal/slice').default;
```
megawac: are these test changes related to the PR?

lib/concatSeries.js
hargasinski: Yup. From the history, it looks like all of the other `doSeries` were converted to `doLimit(<function_name>Limit, 1);` when the `*Series` functions were implemented in terms of `*Limit` [here](https://github.com/caolan/async/commit/acd656ddfda189fadfe4505f2b9e3eae6512f4dc).

mocha_test/groupBy.js
hargasinski: This test was a dupe of the one on [L208](https://github.com/caolan/async/blob/39b20e6da268b2b737df7b068d32620b59011474/mocha_test/groupBy.js#L208). It's under `groupSeries`, so I assume I just forgot to update a copy and pasted test when I originally added the `groupBy` tests.

I was using the `groupBy` tests as a reference for some of the edge cases we usually test.

mocha_test/slice.js
hargasinski: Yes(ish), as `concat` now uses `internal/slice`. I wanted to make sure the behaviour `concat` relies on isn't broken by accident. Plus, `slice` is used in a few of our functions, so it's good to have some tests for it similar to the DLL [ones](https://github.com/caolan/async/blob/master/mocha_test/linked_list.js).

hargasinski:
For what it's worth, an alternate implementation for handling falsy values and preserving order is (not variadic though):
```js
mapLimit(coll, limit, function(val, callback) {
    _iteratee(val, function(err, arr) {
            if (err) return callback(err);
            else if (arguments.length > 1) return callback(null, {result: arr});
            return callback(null);
        });
    }, function(err, mapResults) {
        var result = [];
        for (var i = 0; i < mapResults.length; i++) {
            if (mapResults[i]) {
                result = result.concat(mapResults[i].result);
            }
        }
        return callback(err, result);
    });
}
```
but the performance increase isn't as big as I expected:
```bash
$ node perf/benchmark.js -g concat
Latest tag is  v2.4.1
Comparing v2.4.1 with current on Node v7.10.0
concat(10) v2.4.1 x 72,131 ops/sec ±0.79% (26 runs sampled), 0.0139ms per run
concat(10) current x 56,546 ops/sec ±1.28% (32 runs sampled), 0.0177ms per run
v2.4.1 is faster
concat(300) v2.4.1 x 3,939 ops/sec ±2.57% (30 runs sampled), 0.254ms per run
concat(300) current x 3,157 ops/sec ±1.67% (31 runs sampled), 0.317ms per run
v2.4.1 is faster
concat(10000) v2.4.1 x 9.70 ops/sec ±2.15% (19 runs sampled), 103ms per run
concat(10000) current x 8.66 ops/sec ±3.29% (17 runs sampled), 115ms per run
v2.4.1 is faster
v2.4.1 faster overall (103ms total vs. 116ms total)
v2.4.1 won more benchmarks (3 vs. 0)
```

lib/concatLimit.js
```diff
@@ -14,9 +16,28 @@ import concat from './internal/concat';
...
+        _iteratee(val, function(err /*, ...args*/) {
+            if (err) return callback(err);
+            return callback(null, slice(arguments, 1));
+        });
+    }, function(err, mapResults) {
+        var result = [];
+        var _concat = Array.prototype.concat;
+        for (var i = 0; i < mapResults.length; i++) {
+            if (mapResults[i]) {
+                result = _concat.apply(result, mapResults[i]);
```
hargasinski: One note I just realized about this implementation is that it is breaking for `ArrayLike` objects:
```js
var result = {0: 'foo', 1: 'bar', 2: 'baz', length: 3};`
// previously
[].concat(result); // [{0: 'foo', 1: 'bar', 2: 'baz', length: 3}]
// now
Array.prototype.concat.apply([], args); // ['foo', 'bar', 'baz']
```

aearly:
I've been really busy this past week, will review soon!

lib/concatLimit.js
```diff
@@ -14,9 +16,28 @@ import concat from './internal/concat';
...
+    callback = callback || noop;
+    var _iteratee = wrapAsync(iteratee);
+    mapLimit(coll, limit, function(val, callback) {
+        _iteratee(val, function(err /*, ...args*/) {
+            if (err) return callback(err);
+            return callback(null, slice(arguments, 1));
+        });
+    }, function(err, mapResults) {
+        var result = [];
+        var _concat = Array.prototype.concat;
```
aearly: You can move this var to the top level scope.

lib/concatSeries.js
aearly: Maybe it's time to delete it, then!

mocha_test/concat.js
```diff
@@ -3,92 +3,405 @@ var expect = require('chai').expect;
...
+
+                setTimeout(function() {
+                    next();
+                }, delay);
+            }, function(err, result) {
+                expect(err).to.not.eql(null);
+                expect(result).to.be.an('array').that.is.empty;
+            });
+
+            setTimeout(function() {
```
aearly: Is there another way to test this besides expecting it to be a certain way after a timeout?  This kind of thing leads to flaky tests.

lib/concatLimit.js
hargasinski: Good catch, thank you!

mocha_test/concat.js
hargasinski: Yup, fixed. Thank you! It now uses `async.setImmediate` to wait `arr.length` event loop cycles before checking the result. 

The `*Limit` versions of [`parallel`](https://github.com/caolan/async/blob/master/mocha_test/parallel.js#L320), [`each`](https://github.com/caolan/async/blob/master/mocha_test/each.js#L246), [`map`](https://github.com/caolan/async/blob/master/mocha_test/map.js#L254) and [`groupBy`](https://github.com/caolan/async/blob/master/mocha_test/groupBy.js#L261) use a similar test to the other before the fix. When I get a chance, I'll look into updating them to use `async.setImmediate` too.

hargasinski:
Thanks! Is there anything else that needs to be done for `v2.5`?

aearly:
Nope, just updating the changelog and such.  Want to do that?  I can publish when you're ready.

hargasinski:
@aearly I updated the changelog, I believe the `make` script handles everything else. Thanks! Sorry that it took so long.

aearly:
Excellent, just published v2.5.0!  Thanks!

PR #1466, opened 2017-08-21
Don't memoize results that had an error

### LINKED ISSUE #1465: Memoize on success only
This would be a breaking change.

In today's world, if a memoized function errors on initial call, it will error on all future calls. I would have expected the memoization to only occur on success. This seems like a reasonable assumption given that a typical use-case for an async.memoize would be an external call.

Example case:

```
const getQueueUrl = async.memoize(function(queueName, cb) {
  async.waterfall([
    function(cb) {
      sqs.getQueueUrl(queueName, cb);
    },
    function({ QueueUrl }, cb) {
      cb(null, QueueUrl);
    },
  ], cb);
});
```

If the sqs queue has not been fully initialized by the time `getQueueUrl` is invoked initially, all future invocations of `getQueueUrl` will fail, even after the queue has been fully initialized.

I would be happy to either create a pull request to only memoize on success or update the documentation to clarify that the function will be memoized regardless of success or error.

megawac:
fixes #1465 

This one is debatable if we want to change behaviour in 2.x. I think it makes sense to however could be swayed otherwise. Thoughts @aearly?

### DIFF
--- a/lib/memoize.js
+++ b/lib/memoize.js
@@ -14,6 +14,9 @@ function has(obj, key) {
  * function results against, the callback is omitted from the hash and an
  * optional hash function can be used.
  *
+ * **Note: if the async function errs, the result will not be cached and
+ * subsequent calls will call the wrapped function.**
+ *
  * If no hash function is specified, the first argument is used as a hash key,
  * which may work reasonably if it is a string or a data type that converts to a
  * distinct string. Note that objects and arrays will not behave reasonably.
@@ -63,7 +66,11 @@ export default function memoize(fn, hasher) {
             queues[key] = [callback];
             _fn.apply(null, args.concat(function(/*args*/) {
                 var args = slice(arguments);
-                memo[key] = args;
+                var err = args[0];
+                // #1465 don't memoize if an error occurred
+                if (!err) {
+                    memo[key] = args;
+                }
                 var q = queues[key];
                 delete queues[key];
                 for (var i = 0, l = q.length; i < l; i++) {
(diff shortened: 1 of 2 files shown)


lib/memoize.js
```diff
@@ -63,7 +66,11 @@ export default function memoize(fn, hasher) {
...
                 var args = slice(arguments);
-                memo[key] = args;
+                var err = args[0];
+                // #1465 don't memoize if an error occurred
+                if (!err) {
+                    memo[key] = args;
+                }
                 var q = queues[key];
                 delete queues[key];
                 for (var i = 0, l = q.length; i < l; i++) {
```
hargasinski: We should keep this as is for now, but for v3, I'd be open to changing this behaviour, as there's a potential race condition here (similar to #1248).

If the initial call to the memoized function takes between `m` and `n` seconds to err, and a second call is made `x` seconds after the initial (m < x < n`) then the function might only be run once depending on whether or not the initial call finished. To eliminate it, we would have to call the memoized function again if there are queued calls, and an error occurred.

aearly:
Yes, this is a potential 3.0 change.

It's debate-able as to whether caching errors is "correct" or not, though.  e.g. caching the result of a request that 404's -- do you cache the 404, or keep retrying?

With this change, you could use `reflect` and always cache if you wanted, or use it normally and not cache errors.

aearly:
This is ready to be merged for 3.0.  Also, this new behavior is much less surprising if you were using an `async` function.

Issue #1064, opened 2016-03-15
A solution for stoping a flow after an early exit


Herobs:
There are a strong requirement that stop a flow normally instead of call callback with a fake err. I find a solution from [restify](http://restify.com/):

> You can pass in false to not error, but to stop the handler chain.

So only `null` to indicate a error, and `false` for normal terminal.

`callback(false)` will stop the flow, but won't active error handle like `if (err) console.log(err)`.

aearly:
Hmnn, this is very intriguing!  I like that there is precedent in restify.  It also seems like `false` might be a good to use -- most people pass `null` or `undefined`.  There is a bit of symmetry with how returning `false` in `_.each` also breaks the iteration.

aearly:
@megawac What do you think about this?

megawac:
Works for me (in a major bump) - would be nice for the `createTester`
implementation as it will have an improved method of terminating `*Series`
and `*Limit` based iteration

aearly:
Yeah, I agree, we should save it for the next major.  The breaking change we made with predicate-style functions (e.g. `some`, `filter`) would make this change cause extra confusion as people migrate since people would be passing `false` as the first arg.

wjagodfrey:
Not having this is very inconvenient. I support the use of `cb(false)` - it's what I had expected to exist when first using this module.

megawac:
Think I will be hammering on some of these breaking changes over the weekend.

@hargasinski @aearly any thoughts on a 3.0 timeline? December perhaps?

aearly:
Do we have any stats on how widely used v2 is? I wouldn't want to have more breaking changes for people to evaluate while most people still haven't upgraded.

hargasinski:
We should probably do a minor 2.1 release with the changes applied in #1261 and a non-breaking fix for #1293 before 3.0. The docs will also get updated with the changes in PRs, such as #1263, #1271, and #1280, which will be helpful.

@aearly, that's good point. 2.0 will have been released for 6 months in December. Should we assume that will be enough time for most users to update? We could use the changelog to write a migration guide in the Wiki for 1.5 to 3.0.

megawac:
This is now somewhat possible with `2.1.1` by returning `async/internal/breakLoop`

stephent:
@megawac is there a link to any documentation for this (async/internal/breakLoop)? Not seeing anything on the site so far.

aearly:
[...]

PR #1472, opened 2017-09-16
Feature - add priority race

sandinmyjoints:
This solves a similar problem to https://github.com/caolan/async/issues/1064. For this solution, you can't arbitrarily stop the loop (to have annotate tasks ahead of time), but otoh, it's not a breaking change.

### DIFF
--- a/mocha_test/priorityRace.js
+++ b/mocha_test/priorityRace.js
@@ -1,0 +1,92 @@
+var async = require('../lib');
+var expect = require('chai').expect;
+var assert = require('assert');
+
+describe('priorityRace', function () {
+    it('should call each function in parallel and finish when one prioritized task is done', function priorityRaceTest10(done) {
+        var finished = 0;
+        var tasks = [];
+        function eachTest(i) {
+            var index = i;
+            return function (next) {
+                finished++;
+                next(null, index);
+            };
+        }
+        for (var i = 0; i < 10; i++) {
+            tasks[i] = eachTest(i);
+        }
+        tasks[4].isPrioritized = true
+        async.priorityRace(tasks, function (err, result) {
+            assert.ifError(err);
+            //race ended when 5 finished first
+            expect(result).to.eql([0, 1, 2, 3, 4]);
+            assert.strictEqual(finished, 5);
+            done();
+        });
+    });
+    it('should call each function in parallel and finish when all prioritized tasks are done', function priorityRaceTest20(done) {
+        var finished = 0;
+        var tasks = [];
+        function eachTest(i) {
+            var index = i;
+            return function (next) {
+                finished++;
+                next(null, index);
+            };
+        }
+        for (var i = 0; i < 10; i++) {
+            tasks[i] = eachTest(i);
+        }
+        tasks[0].isPrioritized = true;
+        tasks[7].isPrioritized = true;
+        async.priorityRace(tasks, function (err, result) {
+            assert.ifError(err);
+            //race ended when both 0 and 7 finished
+            expect(result).to.eql([0, 1, 2, 3, 4, 5, 6, 7]);
+            assert.strictEqual(finished, 8);
+            done();
+        });
+    });
+    it('should callback with the first error', function priorityRaceTest30(done) {
+        var tasks = [];
+        function eachTest(i) {
+            var index = i;
+            return function (next) {
+                setTimeout(function () {
+                    next(new Error('ERR' + index));
+                }, 50 - index * 2);
+            };
+        }
+        for (var i = 0; i <= 5; i++) {
+            tasks[i] = eachTest(i);
+            tasks[0].priority = true;
+        }
+        async.priorityRace(tasks, function (err, result) {
+            assert.ok(err);
+            assert.ok(err instanceof Error);
+            assert.strictEqual(typeof result, 'undefined');
+            assert.strictEqual(err.message, 'ERR5');
+            done();
+        });
+    });
+    it('should callback when task is empty', function priorityRaceTest40(done) {
+        async.priorityRace([], function (err, result) {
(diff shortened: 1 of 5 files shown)


sandinmyjoints:
From what I can tell, the Travis failure (for Node 7 only) is unrelated to my changes, maybe a spurious build error?

mocha_test/priorityRace.js
```diff
@@ -0,0 +1,92 @@
...
+            var index = i;
+            return function (next) {
+                setTimeout(function () {
+                    next(new Error('ERR' + index));
+                }, 50 - index * 2);
+            };
+        }
+        for (var i = 0; i <= 5; i++) {
+            tasks[i] = eachTest(i);
+            tasks[0].priority = true;
```
megawac: What is this getting set for?

sandinmyjoints:
@megawac Thanks for reviewing. Great questions. Here's the use case that motivates this. We have two db tables, A and B. Data from A is preferred -- we always use it if available. But not every query has data in A. Some queries have data in B, and if A is not available, we'll use the data from B. 

The simplest thing to do would be:
- query A
- if there's a result, return it
- if no result, query B
- if there's a result, return it, else return no result

That's fine, but we're trying to optimize for speed. Priority race would allow us to kick off queries to A and B concurrently, return as soon as a result is found in A if one is found, else return when either B has a result or neither has a result. This speeds up two of three cases:

| case|time - simple way|time - priority race |
| ----|-----------------|---------------------|
| A has result|time of query A|time of query A |
| A has no result, B has result|time of query A + time of query B| time of the longer of A or B |
| neither has result|time of query A + time of query B| time of the longer of A or B |

So that's the motivation. 

Regarding `.priority = int`, the way I was thinking about things, there are only two classes of tasks: those that cause the entire async operation to finish early vs. those that do not. Hence a boolean. I'm open to something else if you think it makes more sense. To work on an implementation, I'd like to understand what you have in mind. What behavior would you expect with these three tasks:

- task A has priority 0
- task B has priority 1
- task C has priority 2

For these orderings of when each task finishes:
- A, B, C
- A, C, B
- B, A, C
- B, C, A
- C, A, B
- C, B, A

Thanks!

mocha_test/priorityRace.js
sandinmyjoints: To show that when more than one task has priority, the async operation will finish if/when *all* of them finish, not just one of them.

sandinmyjoints:
Also, I agree that assigning a property to functions is a bit odd, certainly uncommon. An alternative is to have some way inside of tasks to signal that the entire async operation should finish before all tasks are done. I read over https://github.com/caolan/async/issues/1064 and it seemed to have a lot of debate about how to accomplish this, so then assigning priority to tasks started to look acceptable -- it sidesteps the question of how tasks can stop the async operation.

mocha_test/priorityRace.js
megawac: Sorry I meant that this should be `isPrioritized`

hargasinski:
@sandinmyjoints could this use case be solved by `async.parallel` and #1064 as opposed to adding a new method? i.e. something along the lines of
```js
async.parallel({
  tableA: (callback) => {
    getDataFromTableA((err, data) => {
      if (err || !data) {
        return callback(err);
      }
      // finish early, don't wait for tableB to finish
      return callback(false, data);
    });
  },
  tableB: (callback) => {
    getDataFromTableB((err, data) => {
      if (err) {
        return callback(err);
      }
      return callback(null, data);
    });
  }
}, (err, results) => {
  if (err) {
    // error handling code
  }

  let data = results.tableA || results.tableB;
  // handle the data
});
```

This has the same performance benefits as `priorityRace`:

case                                       | time - simple way                           | parallel + breaking early
---------------------------------------| ---------------------------------------------| ----------------------------------------------
A has result                            | time of query A                              | time of query A (doesn't wait for B)
A has no result, B has result  | time of query A + time of query B  | time of the longer of A or B
neither has result                    | time of query A + time of query B | time of the longer of A or B

For an int priority, I would expect `priorityRace` to wait for the highest priority to finish. If the highest priority task doesn't return data, it would wait for the next highest priority to finish or if it has already finished, check if it returned data. If it returns/returned data, `priorityRace` would finish. Otherwise, it would continue checking the lower priorities.

mocha_test/priorityRace.js
sandinmyjoints: You're absolutely right, this was left over from an earlier implementation. Fixed.

sandinmyjoints:
@hargasinski I agree that that approach accomplishes the same outcome. My thinking was that calling back with `false` for the first argument is a bit of a workaround -- I know people do it, but it goes against the node convention of the first argument being an error or `null`, and it's not mentioned in the async docs anywhere as a way to accomplish this purpose. My intention with this PR was to make it a first-class use case. Might be personal taste, but it seems like less of a code smell to me to say "set this property on a task to make it short-circuit" than "call back with `false` as the first argument to short-circuit". 😃

aearly:
I feel this method is a bit too niche to add to Async.  It also feels like the control flow can be accomplished with `parallel` and an early return, or a combination of `parallel` and `race`.

I also think the discussion about #1064 is a bit of a red herring.  There's nothing preventing you from calling an outer callback early, but currently, doing so causes a subtle memory leak since the subsequent callbacks wait forever to be called.  The proposed `callback(false)` solution is about signaling that a flow was exited, and no further callbacks will be called, closing the memory leak. (It actually would make control flow like this better! But this control flow is not a solution to that particular problem.)

sandinmyjoints:
OK, thanks for the thoughts and feedback @aearly @hargasinski @megawac. I'll use an alternative approach for my use case.

PR #1483, opened 2017-10-15
Generate index.js from a template

aearly:
I noticed we were leaving out some aliases in our `index.js`.  It actually was not possible to `const async = require('async'); async.find(...)`.

To work around this, I decided to build `lib/index.js` from a template, reading `lib/` and a static `aliases.json` to fill it in.  This also makes maintaining `lib/index.js` easier.  I am also considering not checking it in.

We do not have to have an `aliases.json`, we could also parse source files for `@alias foo` directives in the JSDoc blocks, but want some feedback on that strategy.

We also are missing individual module files for *every* alias.  You cannot `require('async/find')`.  I'm going to add this, once I figure out how to specify it with Make.


megawac:
> We also are missing individual module files for every alias. You cannot require('async/find'). I'm going to add this, once I figure out how to specify it with Make.

I'd rather we punt on this and drop the aliases in v3

aearly:
I think aliases are fine.  And in 2.0, some are actively broken, still needing fixing.

megawac:
I think things like `detect` should definitely be dropped. That alias is very rare and non standard in other similar libs/es2015

aearly:
What do you think about these changes to solve our current bug?

aearly:
I agree about `detect`, `select`, and `inject`, but with the others it's less clear as to which is "correct".

hargasinski:
For adding the missing aliases, this LGTM. I'm not a huge fan of the static `aliases.json` though.

I'm good with dropping less commonly used aliases in v3, but ones like `forEach`, and `forEachOf` we should definitely keep.

Makefile
```diff
@@ -45,6 +47,9 @@ build-bundle: build-modules $(UMD_BUNDLE)
 
 build-modules: $(CJS_MODULES)
 
+$(JS_INDEX): $(INDEX_SRC)
```
hargasinski: Should this be added as a dependency of `build-dist`?

hargasinski: Nevermind, it already is, sorry.

lib/nextTick.js
```diff
@@ -4,7 +4,7 @@ import { hasNextTick, hasSetImmediate, fallback, wrap }  from './internal/setImm
 
 /**
  * Calls `callback` on a later loop around the event loop. In Node.js this just
- * calls `setImmediate`.  In the browser it will use `setImmediate` if
+ * calls `process.nextTicl`.  In the browser it will use `setImmediate` if
```
hargasinski: nit: `process.nextTick`

aearly:
@hargasinski Would you prefer reading the `@alias foo` directives in the doc blocks vs. a static `aliases.json`?

aearly:
Actually, I think `aliases.json` is best, because I can parse/stringify it in such a way that it can become a list of files to generate with Make.  (I could generate it with another script, though...)

hargasinski:
That's a good point. Plus, after thinking about it a little more, the file will rarely change, so it might not even be worth the effort to write another script to generate it. I'm good with these changes.

PR #1507, opened 2017-12-24
Fix NPE when err is undefined

marcelo-mason:
Fixes:
```
TypeError: Cannot read property 'message' of undefined
```


megawac:
Hi @marcelo-mason, good catch! Could you please add a test case please.

aearly:
Re-open this when you have a test case, so we can verify there is a bug here.

PR #1514, opened 2018-02-14
fix "RangeError: Maximum call stack size exceeded"

smbape:
`eachOfLimit` throws a `RangeError: Maximum call stack size exceeded` when the collection is too big and the iteratee is synchronous.
The usual workaround is to use `setImmediate(next)` or [http://caolan.github.io/async/docs.html#ensureAsync](http://caolan.github.io/async/docs.html#ensureAsync).

I have a data analysis program. The `setImmediate(next)` workaround made the whole program slow when there are few new data to analyse.

With a small modification, it was possible to make the program lasting from minutes to seconds.

Here is an exemple of time duration with and without `setImmediate`

```js
const eachOfLimit = require("async/eachOfLimit");

(function() {
    'use strict';

    const collection = new Array(Math.pow(1024, 2));
    for (let i = 0, len = collection.length; i < len; i++) {
        collection[i] = i + 1;
    }

    const timerInit = Date.now();
    eachOfLimit(collection, 1, (value, i, next) => {
        next();
    }, err => {
        console.log("sync", Date.now() - timerInit, "ms");
    });
}());

(function() {
    'use strict';

    const collection = new Array(Math.pow(1024, 2));
    for (let i = 0, len = collection.length; i < len; i++) {
        collection[i] = i + 1;
    }

    const timerInit = Date.now();
    eachOfLimit(collection, 1, (value, i, next) => {
        setImmediate(next);
    }, err => {
        console.log("async", Date.now() - timerInit, "ms");
    });
}());
```

### DIFF
--- a/mocha_test/eachOf.js
+++ b/mocha_test/eachOf.js
@@ -43,6 +43,30 @@ describe("eachOf", function() {
         });
     });
 
+    it('forEachOf no call stack size exceed error', function(done) {
+        var obj = {};
+        var len = 3000;
+        var args = new Array(len * 2);
+        var expected = new Array(len * 2);
+
+        for (var i = 0; i < len; i++) {
+            obj["a" + i] = i;
+            expected[2 * i] = "a" + i;
+            expected[2 * i + 1] = i;
+        }
+
+        async.forEachOf(obj, function(value, key, callback) {
+            var index = parseInt(key.slice(1), 10);
+            args[2 * index] = key;
+            args[2 * index + 1] = value;
+            callback();
+        }, function(err) {
+            assert(err === null, err + " passed instead of 'null'");
+            expect(args).to.eql(expected);
+            done();
+        });
+    });
+
     it('forEachOf - instant resolver', function(done) {
         var args = [];
         async.forEachOf({ a: 1, b: 2 }, function(x, k, cb) {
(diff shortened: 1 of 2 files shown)


mocha_test/eachOf.js
```diff
@@ -274,6 +274,18 @@ describe("eachOf", function() {
...
+    it('forEachOfLimit 1024 * 1024 synchronous call', function(done) {
+        var count = 0;
+        async.forEachOfLimit(_.range(1024 * 1024), Infinity, function(x, i, callback){
+            count++;
+            callback();
+        }, function(err){
+            if (err) throw err;
+            expect(count).to.equal(1024 * 1024);
+        });
+        setTimeout(done, 25);
```
hargasinski: Could you call `done` inside of the final callback instead?

mocha_test/eachOf.js
```diff
@@ -274,6 +274,18 @@ describe("eachOf", function() {
         setTimeout(done, 25);
     });
 
+    it('forEachOfLimit 1024 * 1024 synchronous call', function(done) {
```
hargasinski: This is minor, but we already have a `forEachOfLimit synchronous` test. Could you rename this to something like `forEachOfLimit no stackoverflow` or `forEachOfLimit no call stack size exceed error` just to differentiate between the two?

hargasinski: Could you also add a similar test for `forEachOf` and `forEachOfSeries` to make sure the behaviour is consistent?

hargasinski:
I don't know why the build initially failed, but it appears to be working now. LGTM.

megawac:
👍 I was originally going to comment that this might be a misuse of async, but to be honest I've hit this case in my own personal use enough times to think its justified. That this solution is so simple is just icing on the cake.

Issue #1515, opened 2018-02-27
Return Promise if callback is not provided?


rocketspacer:
Nice to have in ES2017 async/await development
```
const asyncjs = require('async');
const foo = async () => {
    const arr = Array.from({ length: 1000 }, (v, i) => i);
    const transformedArr = await asyncjs.mapLimit(arr, 100, async (v) => {
        return await Promise.resolve(v * 2);
    });
    console.log(transformedArr);
}
```

hargasinski:
Historically, `async` has been against this (see #1086).

However, with the new `async/await` syntax, [node adding promise APIs in the future](https://github.com/nodejs/node/pull/18297), and the support for promises having increased, I'd be open to revisiting this. Thoughts @aearly @megawac?

aearly:
I've thought about this, and experimented with it a bit.  People are going to want to `await Async.mapLimit(arr, 10, async foo => {...})`.  It could very well be something we include in 3.0.

ralphdas:
When using Async I sort of naturally assumed the return of a Promise. It was only at the dis-functioning of my code and my routine rtfm moment that I noticed the callback structure. The above would be very welcome since it would keep code as uniform ES7 as possible.

rorymadden:
I agree that this would be useful. 

In the interim, if anyone is stuck right now, I found this article good for explaining how to write a forEach loop with await. 
https://codeburst.io/javascript-async-await-with-foreach-b6ba62bbf404

aearly:
Finally dug out my old experiment:  #1526

rodrigogs:
+1

tritoanst:
This is my solution:
- Convert to async method
```javascript
function makeAsync() {
  if (arguments.length < 3) return; // wrong params, not run or throw exception here!
  let fixParams = {};
  let paramLength = arguments.length;
  let method = arguments[0];
  for (let i = 1; i < arguments.length; i++) {
    fixParams[(i - 1)] = arguments[i];
  }
  return new Promise((resolve, reject) => {
    const callback = (err, result) => {
      if (err) return reject(err);
      return resolve(result);
    };
    
    fixParams[paramLength - 1] = callback;
    fixParams.length = paramLength;
    method.apply(null, fixParams);
  });
}
```
- and call:
```javascript
await makeAsync(async.each, openFiles, async (file) => {
  await asyncOperation(file);
});
```

ex1st:
@tritoanst, your promisify function is overhead. It may be easier:
[...]

PR #1526, opened 2018-04-15
[wip] POC for await-able Async functions

aearly:
Related to #1515 

This is something I played around with over a year ago.


lib/each.js
```diff
@@ -25,6 +25,7 @@ import wrapAsync from './internal/wrapAsync'
  * If you need the index, use `eachOf`.
  * @param {Function} [callback] - A callback which is called when all
  * `iteratee` functions have finished, or an error occurs. Invoked with (err).
+ * @returns {Promise} a promise, if the callback is omitted
```
aearly: Deliberately never returning anything from an Async function has given us the headroom to do this!

lib/internal/promiseCallback.js
```diff
@@ -0,0 +1,24 @@
+import noop from 'lodash/noop';
+
+var supportsPromise = typeof Promise === 'function';
```
aearly: A pretty crude way to detect promise support.  This would be a 3.0 feature.  I thinking of dropping support for non-ES2015 environments, in which case we could always return a Promise.

mocha_test/es2017/asyncFunctions.js
```diff
@@ -56,11 +65,9 @@ module.exports = function () {
         async.eachOfSeries(input, asyncIdentity, done);
     });
 
-    it('should handle async functions in map', (done) => {
-        async.map(input, asyncIdentity, (err, result) => {
-            expect(result).to.eql(input);
-            done(err);
-        });
+    it('should handle async functions in map', async () => {
+        var result = await async.map(input, asyncIdentity);
```
aearly: :heart_eyes:

mocha_test/es2017/asyncFunctions.js
```diff
@@ -32,8 +33,16 @@ module.exports = function () {
      * Collections
      */
 
-    it('should handle async functions in each', (done) => {
-        async.each(input, asyncIdentity, done);
+    it('should handle async functions in each', async () => {
+        var promise = async.each(input, asyncIdentity);
+        assert(typeof promise.then === 'function');
```
ex1st: Why not simpler?

```javascript
assert(promise instanceof Promise);

```

mocha_test/es2017/asyncFunctions.js
```diff
@@ -32,8 +33,16 @@ module.exports = function () {
...
+        assert(Object.getPrototypeOf(promise) === promiseProto);
```
ex1st: Why not simpler?

```javascript
assert(promise instanceof Promise);
```

lib/internal/promiseCallback.js
hargasinski: :+1: for dropping pre-ES2015 support in v3.

hargasinski:
This is exciting! Looks good so far, let me know if you need help with this PR though. I could add the rest of the tests/docs.

lib/internal/once.js
```diff
@@ -1,8 +1,10 @@
 export default function once(fn) {
-    return function () {
+    function wrapped () {
         if (fn === null) return;
         var callFn = fn;
         fn = null;
         callFn.apply(this, arguments);
     };
+    wrapped.promise = fn.promise;
```
megawac: Should the wrapped function not return a promise instead to allow `wrapped().then(...)` or how do we intend this to work?

lib/internal/eachOfLimit.js
```diff
@@ -50,5 +50,6 @@ export default function _eachOfLimit(limit) {
         }
 
         replenish();
+        return callback.promise;
```
megawac: I think we'll want to track whether the callback is a user defined callback or an internal promise callback and only return `promiseCallback.promise` if its an internal version to avoid cases where users pass us a callback with some state on it.

DanielRamosAcosta:
Hi Guys!

I've just saw this PR, and I want to tell you guys I've been developing an alternative version of `async` with promises. Here is the repo: [DanielRamosAcosta/async-promised](https://github.com/DanielRamosAcosta/async-promised).

Currently I’ve promisified about 65% of the library functions (with their respective tests), you can see the progress here: https://github.com/DanielRamosAcosta/async-promised/blob/master/TODO.md

My plans were:

* Implement all the test with async/await/promises
* Do it just wrapping the original async library
* Fix docs & TypeScript typings
* Implement all functions in a promise-native way, so I can drop  the original async dependency
 
But now that I’ve seen this PR, maybe we can do something together! I’ll appreciate any thoughts you share with me.

aearly:
This branch has diverged greatly from master, I intend to re-implement this feature soon.

aearly:
Closing in favor of #1572

PR #1529, opened 2018-04-22
allow methods in waterfall to be called by a specific caller

kunal361:


hargasinski:
Thanks for the PR!

However, I'm not convinced this belongs in `async` as JavaScript already has a few of ways to bind a specific context to a function:

```js
// from the tests added

// bind
async.waterfall([
  test.method1.bind(test)
], function(err, results) {
  // ...
});

// anonymous function
async.waterfall([
  (cb) => test.method1(cb)
], function(err, results) {
  // ...
});
```

kunal361:
@hargasinski Agreed.
But my thought behind this PR was that there has to be a way to utilise `apply`'s first argument.

`bind` creates a new function so not really sure if it is necessary and/or optimal.

And for the anonymous function, would have to list down all the arguments if there were any, which seems to be extra bit of writing.

Not sure if i am missing something. Also thanks for taking out time to review this.

aearly:
I agree with @hargasinski here.   We have historically left `this`-binding of functions to the user as well.

The performance of `.bind()` and arrow functions has also been greatly improved in recent versions of V8 as well, I wouldn't worry about it.

PR #1531, opened 2018-05-08
moved initialization of index into for loop expression

peterhendrick:
Was getting an exception on this line, and this prevented the exception.

Also makes it more readable.

### DIFF
--- a/lib/eachOf.js
+++ b/lib/eachOf.js
@@ -11,8 +11,7 @@ import wrapAsync from './internal/wrapAsync';
 // eachOf implementation optimized for array-likes
 function eachOfArrayLike(coll, iteratee, callback) {
     callback = once(callback || noop);
-    var index = 0,
-        completed = 0,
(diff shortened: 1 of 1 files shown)


aearly:
Do you have an example failing test case?

By my eyes, this change should have no effect on how it actually executes.

peterhendrick:
I agree, it shouldn't change any behavior based on the look, however, I was running a script that was using async (node v 8.11.1), and my IDE was breaking on this line of the for expression (it breaks on exceptions).

When I ran the script with this change, it stopped breaking there. (sorry can't share the script).

I figured there must be something weird going on, and looking at the code, this change at least made it more readable, and it stopped breaking. But as to why there was no longer an exception, I really don't know.

aearly:
Closing this, since it doesn't appear to have an effect.  If setting a breakpoint there fixed an issue, it sounds like a race condition or timing issue.

PR #1528, opened 2018-04-19
Drop lodash

bfred-it:
Fixes #1283 

:)

PR #1533, opened 2018-05-16
Update lodash

### LINKED ISSUE #1532: Update or drop lodash due to Vulnerability
The latest version of npm has  a new audit feature to expose potential vulnerabilities. There is currently a low severity vulnerability in lodash. 

The issue is patched in lodash **v4.17.5+**

Async is currently on lodash **v4.14.0** (https://github.com/caolan/async/blob/master/package.json#L22)

More details on https://nodesecurity.io/advisories/577

I see there are also plans to drop lodash. That would fix this as well.

chrisdukakis:
Fixes #1532 . Updates lodash in all deps, in light of https://nodesecurity.io/advisories/577, just in case.
Pls review and merge this or #1528 , we all need it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "utility"
   ],
   "dependencies": {
-    "lodash": "^4.14.0",
+    "lodash": "^4.17.5",
     "lodash-es": "^4.14.0"
   },
   "devDependencies": {
@@ -48,7 +48,7 @@
     "karma-mocha-reporter": "^2.2.0",
     "mocha": "^3.1.2",
     "native-promise-only": "^0.8.0-a",
-    "nyc": "^7.0.0",
+    "nyc": "^11.8.0",
     "recursive-readdir": "^1.3.0",
     "rimraf": "^2.5.0",
     "rollup": "^0.36.3",
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
   ],
   "dependencies": {
     "lodash": "^4.17.5",
-    "lodash-es": "^4.14.0"
+    "lodash-es": "^4.17.10"
   },
   "devDependencies": {
     "babel-cli": "^6.24.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -19,7 +19,7 @@
     "utility"
   ],
   "dependencies": {
-    "lodash": "^4.14.0",
+    "lodash": "^4.17.5",
     "lodash-es": "^4.14.0"
```
hargasinski: Does `lodash-es` need to be updated too?

package.json
```diff
@@ -48,7 +48,7 @@
     "karma-mocha-reporter": "^2.2.0",
     "mocha": "^3.1.2",
     "native-promise-only": "^0.8.0-a",
-    "nyc": "^7.0.0",
+    "nyc": "^11.8.0",
```
hargasinski: I'm assuming this is because the old version of `nyc` was using an older version of `lodash`?

hargasinski:
Thanks for the PR! I just added a couple of quick comments.

I'm okay with merging this for a quick patch release until #1528 gets merged and released in `v3`.

package.json
chrisdukakis: that's correct

package.json
chrisdukakis: `lodash-es` doesn't depend on `lodash`, it's fine as is

chrisdukakis:
Thank you @hargasinski ! I only updated packages using lodash, there are more cases that you might want to look into, but all of those are dev deps. I wasn't familiar with those tools, so I didn't push further changes and brake anything. You can try an `npm audit` and check the log.

package.json
hargasinski: As far as I know, `lodash-es` is built from `lodash`. 

Either way, it's probably a good idea to keep them synced. Could you update it to `v4.17.5`? The PR will be good to go afterwards.

chrisdukakis: You are right! Done!

aearly:
Going to merge this for a patch, will remove Lodash in 3.0

PR #1542, opened 2018-06-05
feat: Canceling flows

### LINKED ISSUE #1064: A solution for stoping a flow after an early exit

aearly:
Closes #1064 

This allows you to signal that you have exited early from an async flow by `callback(false)` --- calling back with `false` as the error.  All further callbacks will be ignored, and the final callback will not be called.  This is handy to avoid a subtle memory leak when you exit early.

- [x] `eachOfLimit` - based methods (e.g. all collection-based methods)
- [x] `compose`/`seq`
- [x] `waterfall`
- [x] `auto`/`autoInject`
- [x] `whilst`, `until`, and family
- [x] `forever`
- [ ] ?? (any I'm missing?)
- [x] Docs

### DIFF
--- a/intro.md
+++ b/intro.md
@@ -175,7 +175,7 @@ async.map([1, 2, 3], AsyncSquaringLibrary.square.bind(AsyncSquaringLibrary), fun
 
 ### Subtle Memory Leaks
 
-There are cases where you might want to exit early from async flow, when calling an Async method insice another async function:
+There are cases where you might want to exit early from async flow, when calling an Async method inside another async function:
 
 ```javascript
 function myFunction (args, outerCallback) {
--- a/intro.md
+++ b/intro.md
@@ -173,6 +173,40 @@ async.map([1, 2, 3], AsyncSquaringLibrary.square.bind(AsyncSquaringLibrary), fun
 });
 ```
 
+### Subtle Memory Leaks
+
+There are cases where you might want to exit early from async flow, when calling an Async method insice another async function:
+
+```javascript
+function myFunction (args, outerCallback) {
+    async.waterfall([
+        //...
+        function (arg, next) {
+            if (someImportantCondition()) {
+                return outerCallback(null)
+            }
+        },
+        function (arg, next) {/*...*/}
+    ], function done (err) {
+        //...
+    })
+}
+```
+
+Something happened in a waterfall where you want to skip the rest of the execution, so you call an outer callack.  However, Async will still wait for that inner `next` callback to be called, leaving some closure scope allocated.
+
+As of version 3.0, you can call any Async callback with `false` as the `error` argument, and the rest of the execution of the Async method will be stopped or ignored.
+
+```javascript
+        function (arg, next) {
+            if (someImportantCondition()) {
+                outerCallback(null)
+                return next(false) // ← signal that you called an outer callback
+            }
+        },
+```
+
+
 ## Download
 
 The source is available for download from
--- a/lib/auto.js
+++ b/lib/auto.js
@@ -102,6 +102,7 @@ export default function (tasks, concurrency, callback) {
 
     var results = {};
     var runningTasks = 0;
+    var canceled = false;
     var hasError = false;
 
     var listeners = Object.create(null);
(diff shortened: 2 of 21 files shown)


aearly:
This is ready!

intro.md
```diff
@@ -173,6 +173,40 @@ async.map([1, 2, 3], AsyncSquaringLibrary.square.bind(AsyncSquaringLibrary), fun
 });
 ```
 
+### Subtle Memory Leaks
+
+There are cases where you might want to exit early from async flow, when calling an Async method insice another async function:
```
megawac: Typo `-insice/+inside`

lib/internal/eachOfLimit.js
```diff
@@ -14,17 +14,23 @@ export default function _eachOfLimit(limit) {
...
                 done = true;
                 callback(err);
             }
+            else if (err === false) {
+                done = true;
+                canceled = true;
+            }
             else if (value === breakLoop || (done && running <= 0)) {
                 done = true;
+                if (canceled) return
```
megawac: Move to top of `iterateeCallback` for simplicity? (it would also let you avoid the `!canceled` check in the error condition)

lib/auto.js
```diff
@@ -102,6 +102,7 @@ export default function (tasks, concurrency, callback) {
 
     var results = {};
     var runningTasks = 0;
+    var canceled = false;
```
megawac: Typo should be `cancelled`

aearly: Either spelling is correct.  I chose one 'l' for consistency.

test/during.js
```diff
@@ -34,6 +34,22 @@ describe('during', function() {
         );
     });
 
+    it('during canceling', (done) => {
```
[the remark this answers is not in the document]
aearly: We're going to deprecate during, making all whilst/until functions have an async test function, so no need for comprehensive test coverage.

lib/auto.js
megawac: damn americans

aearly:
Ok, I believe I've addressed all your concerns. Thanks for the review!

And yes, I believe not calling the final callback is what we want here, as a mechanism to avoid memory leaks.  In what cases would you want to break iteration, and have the final callback called?

test/auto.js
```diff
@@ -168,6 +168,63 @@ describe('auto', function () {
...
+        });
+        setTimeout(() => {
+            expect(call_order).to.eql([1, 3])
+            done()
+        }, 10);
+    });
+
+    it('does not start other tasks when it has been canceled', function(done) {
+        const call_order = []
+        debugger
```
megawac: Misc debugger statement

hargasinski:
That's a good point @aearly. Thanks for making the changes, lgtm!

nathaniel-holder:
> Ok, I believe I've addressed all your concerns. Thanks for the review!
> 
> And yes, I believe not calling the final callback is what we want here, as a mechanism to avoid memory leaks. In what cases would you want to break iteration, and have the final callback called?

@aearly In my code, the final callback of a waterfall calls the parent callback, so it would break my existing code to use this cancel feature.  For example:
```
function handler(params, done) {
    async.waterfall(
        [
            function (callback) {
                console.log('step 1');
                return callback();
            },
            function (callback) {
                console.log('step 2');
                // skip remaining steps
                return callback(false);
            },
            function (callback) {
                console.log('step 3');
                return callback();
            }
        ],
        function (err) {
            console.log('final callback');
            if (err) return done(err);
            return done(null, 'data');
        }
    );
}
handler(null, (err, data) => console.log('This will never be called. err: %j, data: %j', err, data));
```
Is this not a use case for others?

yeya:
This feature breaks the basic node API, passing false to the callback should not be treated differently than `null`, `undefined`, or any other falsy item.
It's not enough to add a line in the docs, sometimes you pass functions you didn't write by yourself.
Is there an option to cancel this behavior?

meetSuraj:
@aearly Can you take a look at @nathaniel-holder and @yeya comment?

PR #1553, opened 2018-07-02
ES6-ify codebase

### LINKED ISSUE #1541: Go full ES6 in v3.0?
Currently, the only ES2015 feature we use is modules, which we compile away down to ES5 in the build process.  As we are gearing up for another batch of breaking changes in the 3.0 release, we could also start using more ES6 features.

Node 4 just left its maintenance period,  meaning we are left with Node 6, 8, and 10, which all have nearly full ES6 support.  The situation for all major browsers is similar.  https://kangax.github.io/compat-table/es6/

Pros:
- Better developer ergonomics, smaller source code size
- Lots of ES6 features (e.g. arrow functions, rest parameters) are faster than their ES5 equivalents
- Can use Maps, Sets, Symbols, generators, Promises, etc.  which are all handy

Cons:
- Users still supporting ES5 environments, e.g. IE11 will have to transpile, which will slow down their builds.  They can elect to keep using v2 however.

aearly:
Lots of arrows and rest-params!  I also got rid of some internal functions that were either no longer needed (`slice`), or better inlined (`forOwn`, `notId`, etc.)

Closes #1541


megawac:
Awesome! I'm surprised you didn't have to touch the eslintrc.

How is the performance before and after making these changes? I'm most curious about the performance of `restParams` vs ES6 rest params and about `slice` vs ES6 `spread`

hargasinski:
There's so many big changes coming in, this is awesome! LGTM.

aearly:
In Node 8 and 10, the benchmarks were about the same.  No method was significantly faster or slower than another.

PR #1557, opened 2018-07-09
breaking: remove during, make test functions in until/whilst async

### LINKED ISSUE #850: Add async test func support for whilst/until, deprecate during?
Alright, few qualms with `during` and `doDuring`, along with the synchronous `whilst`, `doWhilst`, `until` and `doUntil`.

First of all, how is `during` semantically correct? _During_ means _within a duration_, notwithstanding iteration (but instead, time).

Secondly, why was `during` and its sister function chosen over an `Function.length === 0` approach for two identical methods `whilst` and `doWhilst` (and then left out for the two counterparts, `until` and `doUntil`)?

Third, doesn't a synchronous test completely go against the idea of Async to begin with?

---

Why not
- **remove**/**deprecate** `during` and `doDuring` (i.e. revert #800)
- **deprecate** the synchronous test forms of `whilst`, `doWhilst`, `until` and `doUntil`
- add non-breaking functionality to those four methods by performing a `Function.length` check on the passed function and treating it as async if its length is greater than 0
- ultimately phase out the use of the synchronous check

These methods are severely crippled by breaking the most obvious pattern of this library, and adding copies to fix their problems is a messy solution.

I'd be happy to do a pull request.

aearly:
Closes #850


lib/until.js
```diff
@@ -13,15 +14,30 @@ import whilst from './whilst';
  * @method
  * @see [async.whilst]{@link module:ControlFlow.whilst}
  * @category Control Flow
- * @param {Function} test - synchronous truth test to perform before each
- * execution of `iteratee`. Invoked with ().
+ * @param {Function} test - asynchronous truth test to perform before each
```
hargasinski: nit: I think the parameter type is `AsyncFunction`.

lib/doUntil.js
```diff
@@ -12,14 +13,18 @@ import doWhilst from './doWhilst';
  * @category Control Flow
  * @param {AsyncFunction} iteratee - An async function which is called each time
  * `test` fails. Invoked with (callback).
- * @param {Function} test - synchronous truth test to perform after each
- * execution of `iteratee`. Invoked with any non-error callback results of
- * `iteratee`.
+ * @param {Function} test - asynchronous truth test to perform after each
```
hargasinski: nit: Sorry, same as before, I think the param type is `AsyncFunction`.

Qix-:
:clap: :feelsgood:

PR #1558, opened 2018-07-09
Add limit value to `mapLimit` example

abraham:


aearly:
Ooh, good catch, thanks!

PR #1560, opened 2018-07-10
feat: initial async generator support

### LINKED ISSUE #1551: Does not work with async generators
Seems so.

aearly:
Closes #1551

We have to create a special `eachOfLimit` implementation for this, but once fully tested, all collections methods should fall into place.

### DIFF
--- a/lib/internal/asyncEachOfLimit.js
+++ b/lib/internal/asyncEachOfLimit.js
@@ -10,15 +10,17 @@ export default function asyncEachOfLimit(generator, limit, iteratee, callback) {
 
     function replenish() {
         //console.log('replenish')
-        if (running >= limit || awaiting) return
+        if (running >= limit || awaiting || done) return
         //console.log('replenish awaiting')
         awaiting = true
         generator.next().then(({value, done: iterDone}) => {
             //console.log('got value', value)
+            if (canceled || done) return
             awaiting = false
             if (iterDone) {
                 done = true;
                 if (running <= 0) {
+                    //console.log('done nextCb')
                     callback(null)
                 }
                 return;
@@ -32,24 +34,26 @@ export default function asyncEachOfLimit(generator, limit, iteratee, callback) {
 
     function iterateeCallback(err, result) {
         //console.log('iterateeCallback')
-        if (canceled) return
         running -= 1;
+        if (canceled) return
         if (err) return handleError(err)
 
         if (err === false) {
             done = true;
             canceled = true;
+            return
         }
 
         if (result === breakLoop || (done && running <= 0)) {
             done = true;
-            //console.log('done')
+            //console.log('done iterCb')
             return callback(null);
         }
         replenish()
     }
 
     function handleError(err) {
+        if (canceled) return
         awaiting = false
         done = true
         callback(err)
--- a/lib/internal/asyncEachOfLimit.js
+++ b/lib/internal/asyncEachOfLimit.js
@@ -1,4 +1,4 @@
-const breakLoop = require('./breakLoop')
+import breakLoop from './breakLoop';
 
 // for async generators
 export default function asyncEachOfLimit(generator, limit, iteratee, callback) {
--- a/lib/internal/asyncEachOfLimit.js
+++ b/lib/internal/asyncEachOfLimit.js
@@ -1,0 +1,59 @@
+const breakLoop = require('./breakLoop')
+
+// for async generators
+export default function asyncEachOfLimit(generator, limit, iteratee, callback) {
+    let done = false
+    let canceled = false
+    let awaiting = false
+    let running = 0
+    let idx = 0
+
+    function replenish() {
+        //console.log('replenish')
+        if (running >= limit || awaiting) return
+        //console.log('replenish awaiting')
+        awaiting = true
+        generator.next().then(({value, done: iterDone}) => {
+            //console.log('got value', value)
+            awaiting = false
+            if (iterDone) {
+                done = true;
+                if (running <= 0) {
+                    callback(null)
+                }
+                return;
+            }
+            running++
+            iteratee(value, idx, iterateeCallback)
+            idx++
+            replenish()
+        }).catch(handleError)
+    }
+
+    function iterateeCallback(err, result) {
+        //console.log('iterateeCallback')
+        if (canceled) return
+        running -= 1;
+        if (err) return handleError(err)
+
+        if (err === false) {
+            done = true;
+            canceled = true;
+        }
+
+        if (result === breakLoop || (done && running <= 0)) {
+            done = true;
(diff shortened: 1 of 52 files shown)


lib/internal/asyncEachOfLimit.js
```diff
@@ -0,0 +1,59 @@
...
+        }).catch(handleError)
+    }
+
+    function iterateeCallback(err, result) {
+        //console.log('iterateeCallback')
+        if (canceled) return
+        running -= 1;
+        if (err) return handleError(err)
+
+        if (err === false) {
```
megawac: Can we avoid calling replenish int his case and go straight to calling the complete callback?

lib/internal/asyncEachOfLimit.js
```diff
@@ -0,0 +1,59 @@
...
+export default function asyncEachOfLimit(generator, limit, iteratee, callback) {
+    let done = false
+    let canceled = false
+    let awaiting = false
+    let running = 0
+    let idx = 0
+
+    function replenish() {
+        //console.log('replenish')
+        if (running >= limit || awaiting) return
```
megawac: How do you end up with more than one generator running asynchronously with this code? If I understand the code correctly, `replenish` only gets called

1) to spawn the generator when `asyncEachOfLimit` is called
2) whenever a generator completes
3) after `iterateeCallback` is called

Do we not need to loop `limit - running` inside of replenish to call several generators at once?

lib/internal/asyncEachOfLimit.js
aearly: You're right, I should `return` in this block.

lib/internal/asyncEachOfLimit.js
aearly: You'd only end up with more than one iteratee running if your iteratee takes longer to process than your generator takes to generate.   For example, in the tests, I have the generator ticking every 1 ms, but have the iteratee take 5 ms.

Async generators are like streams -- you get a linear sequence of items.  You can't await multiple items at the same time (and if you did, you'd get the same item multiple times, as you do when you call `.then()` on the same promise multiple times).  This is why we can't loop in `replenish` -- we'd end up calling the iteratee with the same item multiple times.

megawac: I see thanks for explaining, I haven't used generators before.

So for my understanding, the replenish inside of the generator's `then` function is to spawn the asynchronous processes up until `limit` are running and the `replenish` inside of the `iterateeCallback` is to trigger additional items to run if the limit was reached?

lib/internal/asyncEachOfLimit.js
```diff
@@ -0,0 +1,59 @@
...
+        //console.log('replenish awaiting')
+        awaiting = true
+        generator.next().then(({value, done: iterDone}) => {
```
megawac: Can you have a generator that has no items?

megawac: Nevermind dumb question, I see how this works

lib/internal/asyncEachOfLimit.js
```diff
@@ -0,0 +1,59 @@
...
+            done = true;
+            canceled = true;
+        }
+
+        if (result === breakLoop || (done && running <= 0)) {
+            done = true;
+            //console.log('done')
+            return callback(null);
+        }
+        replenish()
```
megawac: I think you have to avoid calling `replenish` when `done === true` or have replenish check if `done === true`.

Lets say hypothetically we have a limit of 2 and the final 2 `iteratees` are running. When one of these iteratees resolves, `running = 1` and `done = true` but we'll still replenish and call the `generator.next().then()`. This has the possibility of entering a race condition, if the other iteratee resolves in the time between the `then` callback is called, you can actually end up calling `callback` twice

lib/internal/asyncEachOfLimit.js
megawac: > You can't await multiple items at the same time (and if you did, you'd get the same item multiple times...)

Are you sure of this, as is right now we have the possibility of calling `next` multiple times before the promise resolves if `iteratee` is synchronous. In such a case will things break?

hargasinski: > This is why we can't loop in replenish -- we'd end up calling the iteratee with the same item multiple times.

Sorry, I may be misunderstanding something, but I thought calls to `next` are queued and resolved separately. I'm looking at the `Async iterators and async iterables` section of the [proposal](https://github.com/tc39/proposal-async-iteration#async-iterators-and-async-iterables).

That being said, I think we should still probably avoid looping in replenish as we still have to wait for earlier promises to resolve.

lib/internal/asyncEachOfLimit.js
hargasinski: Similar to what @megawac was saying, I think we need a `if (done) return;` line at the very top of this callback. If one of the previous iteratees cancelled or errored, we shouldn't be invoking the` iteratee` with successive items.

For example, if the first `iteratee` starts and we start waiting for the second item. If the first `iteratee` cancels before the `then` callback is called with the second item, when it is called, we would still invoke the `iteratee` with the second item.

lib/internal/eachOfLimit.js
```diff
@@ -15,6 +17,9 @@ export default (limit) => {
         if (!obj) {
             return callback(null);
         }
+        if (isAsyncGenerator(obj)) {
```
hargasinski: This is a minor thing, but should this be expanded to support async iterators in general? Something similar to how we currently have `getIterator` except with `Symbol.asyncIterator`.

lib/internal/asyncEachOfLimit.js
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    function handleError(err) {
```
hargasinski: I believe we need to handle the canceled case here too, as this is the `catch` callback.

lib/internal/asyncEachOfLimit.js
aearly: Ah, you're right, I was confusing `.next()` with `.then()`.  We can call `.next()` as many times as we need to, perhaps up to the concurrency limit.

lib/internal/eachOfLimit.js
aearly: Ah, good point, didnt know about that detail of the spec.

megawac:
@aearly can we update our eslint config to be consistent on semi colons please

aearly:
Okay, I've added support for async iterables.  Implementation was pretty simple, hard part is getting all the various babel stuff to be happy.

aearly:
Also, I'd want to do a linter change in a separate PR.  Going to be a big ugly diff if we switch.

PR #1562, opened 2018-07-14
add item except error happend.

yujunlong2000:
If error happened, the result with a undefined at last. For example dir2 is not exist.
`async.concat(['dir1','dir2'], fs.readdir, function(err, results) {
    // results is now an array of stats for each file
});`


aearly:
Sorry, but we want to keep this behavior. Some libraries will callback with an error and a result in certain cases.

PR #1564, opened 2018-07-20
Updated outdated documentation for Concat

alexandercerutti:
Concat function, since v2.5.0, returns back array results in the original order.


aearly:
Thanks!

PR #1569, opened 2018-08-02
[issue-1568][bug]: make sure error object defined before access its message

### LINKED ISSUE #1568: Make sure error object is defined before access its message
https://github.com/caolan/async/blob/master/lib/asyncify.js#L90
``` javascript
function handlePromise(promise, callback) {
    return promise.then(value => {
        invokeCallback(callback, null, value);
    }, err => {
        invokeCallback(callback, err.message ? err : new Error(err));
    });
}
```
`err && err.message` will be better when I happened to reject a promise without reason.

reedfeng:
fix #1568

### DIFF
--- a/lib/asyncify.js
+++ b/lib/asyncify.js
@@ -87,7 +87,7 @@ function handlePromise(promise, callback) {
     return promise.then(value => {
         invokeCallback(callback, null, value);
     }, err => {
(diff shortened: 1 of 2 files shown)


aearly:
Thanks!

Issue #956, opened 2015-11-13
Promise support


rumkin:
Add promise support to allow mix different tasks:

```javascript
async.waterfall([
    function (done) {
        // Do something and call done()
        fs.readFile(filepath, done);
    },
    function(content) {
        // Do something and return Promise
        return mongoose.models.file.create({
            name: filepath,
            content: content
        });
    }
], (err) => {
    if (err) {
        console.error(err);
    }
});
```

aearly:
`asyncify` will take a synchronous function that returns a Promise and call a callback on its resolved/rejected handlers:

```js
async.waterfall([
    function (done) {
        // Do something and call done()
        fs.readFile(filepath, done);
    },
    async.asyncify(function(content) {
        // Do something and return Promise
        return mongoose.models.file.create({
            name: filepath,
            content: content
        });
    })
], (err) => {
    if (err) {
        console.error(err);
    }
});
```

But that feature isn't documented...

aearly:
I also had another thought -- should we automatically asyncify functions that return promises (or handle them appropriately?).  What should async do when passed an ES-whatever `async` function (that implicitly returns a promise)?

e.g.

```js
async.waterfall([
    function (done) {
        // Do something and call done()
        fs.readFile(filepath, done);
    },
    async function(content) {
        return await mongoose.models.file.create({
            name: filepath,
            content: content
        });
    }
], (err, model) => {
    if (err) {
        console.error(err);
    }
});
```

SEAPUNK:
Personally, I think async should asyncify promises by default. There are some async functions that I write that I also have to feed into async.queue, but I don't want to write this:

```javascript
import {queue, asyncify} from 'async'

const run = asyncify(async function () {
  await someStuff()
})

const q = async.queue(run)
q.push('asdf')
```

where I could be writing this

```javascript
import {queue} from 'async'

async function run () {
  await someStuff()
}

const q = async.queue(run)
q.push('asdf')
```

aearly:
Added the docs for `asyncify`.  Going to keep this open for the automatic asyncify-ing behavior.

Kikobeats:
About this, I was experimenting with use the same codebase and using a Promise interface for methods that use callback as last parameter. [check it](https://github.com/Kikobeats/promise-async/blob/master/index.js) !

```js
async.waterfall([
  function (callback) {
    callback(null, 'one', 'two')
  },
  function (arg1, arg2, callback) {
[...]

Issue #439, opened 2014-02-04
promise support


khomyakov42:
Please add promise support

PR #1572, opened 2018-08-05
feat: await-able Async methods

### LINKED ISSUE #1515: Return Promise if callback is not provided?

aearly:
Closes #1515

### DIFF
--- a/lib/internal/awaitify.js
+++ b/lib/internal/awaitify.js
@@ -1,7 +1,7 @@
 // conditionally promisify a function.
 // only return a promise if a callback is omitted
-export default function awaitify (asyncFn, arity) {
-    if (arity == null) throw new Error('arity is undefined')
+export default function awaitify (asyncFn, arity = asyncFn.length) {
+    if (!arity) throw new Error('arity is undefined')
     function awaitable (...args) {
         if (typeof args[arity - 1] === 'function') {
             return asyncFn.apply(this, args)
--- a/lib/internal/awaitify.js
+++ b/lib/internal/awaitify.js
@@ -1,16 +1,17 @@
 // conditionally promisify a function.
 // only return a promise if a callback is omitted
 export default function awaitify (asyncFn, arity) {
-    const awaitable = function (...args) {
-        if (args.length === arity || typeof args[arity - 1] === 'function') {
+    if (arity == null) throw new Error('arity is undefined')
+    function awaitable (...args) {
+        if (typeof args[arity - 1] === 'function') {
             return asyncFn.apply(this, args)
         }
 
         return new Promise((resolve, reject) => {
-            args.push((err, ...cbArgs) => {
+            args[arity - 1] = (err, ...cbArgs) => {
                 if (err) return reject(err)
                 resolve(cbArgs.length > 1 ? cbArgs : cbArgs[0])
-            })
+            }
             asyncFn.apply(this, args)
         })
     }
--- a/lib/internal/awaitify.js
+++ b/lib/internal/awaitify.js
@@ -1,0 +1,25 @@
+export const ASYNC_FN = Symbol('asyncFunction')
+
+// conditionally promisify a function.
+// only return a promise if a callback is omitted
+export default function awaitify (asyncFn, arity) {
+    function awaitable(...args) {
+        if (args.length === arity || typeof args[arity - 1] === 'function') {
+            return asyncFn.apply(this, args)
+        }
+
+        return new Promise((resolve, reject) => {
+            args.push((err, ...cbArgs) => {
+                if (err) return reject(err)
+                resolve(cbArgs.length > 1 ? cbArgs : cbArgs[0])
+            })
+            asyncFn.apply(this, args)
+        })
+    }
+
+    awaitable[Symbol.toStringTag] = 'AsyncFunction'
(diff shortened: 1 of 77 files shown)


lib/internal/promiseCallback.js
```diff
@@ -0,0 +1,19 @@
+const PROMISE_SYMBOL = Symbol('promiseCallback')
+
+function promiseCallback () {
```
aearly: This is the older strategy.  It's a bit more clunky, but doesn't add another frame to the call stack for each method.

lib/internal/awaitify.js
```diff
@@ -0,0 +1,25 @@
...
+        return new Promise((resolve, reject) => {
+            args.push((err, ...cbArgs) => {
+                if (err) return reject(err)
+                resolve(cbArgs.length > 1 ? cbArgs : cbArgs[0])
+            })
+            asyncFn.apply(this, args)
+        })
+    }
+
+    awaitable[Symbol.toStringTag] = 'AsyncFunction'
```
aearly: This is a bit of a hack, not sure if we should do this.

lib/internal/awaitify.js
```diff
@@ -0,0 +1,25 @@
...
+    awaitable[ASYNC_FN] = true
```
aearly: I intend to use this so we can avoid an additional layer of wrapping in `wrapAsync`.

Themayu:
> Or we just punt on this completely. You can't await wrapped functions. This would be a bummer, because `await timeout(myFn, 1000)(...args)` is pretty powerful.

How exactly would this be powerful?
(I just want an explanation because I don't really understand myself)

aearly:
> How exactly would this be powerful?

It's powerful for the same reason async/await is powerful. `await myFn(...args)` is pretty powerful, so being able to seamlessly slip in a timeout wrapper (or memoize wrapper, or reflect) is even more powerful.

lib/internal/promiseCallback.js
aearly: I had to use this in a couple places where the `awaitify` wrapper couldn't work  (usually due to optional arguments).

Themayu:
Ah. Alright then

aearly:
@megawac @hargasinski I would love to get a review on this before merging.  In regards to my above comment about the wrapper Utils methods, I'm thinking we can defer a solution there until later (temporary punt).

hargasinski:
@aearly really sorry for the delay. These past few weeks have busy for me, and the PR ended up getting drowned out by other things. Thanks for the reminder! I will review it this weekend. I want to spend some time playing around with it, just to make sure I understand it fully.

lib/internal/once.js
```diff
@@ -1,8 +1,10 @@
 export default function once(fn) {
-    return function (...args) {
+    function wrapper (...args) {
         if (fn === null) return;
         var callFn = fn;
         fn = null;
         callFn.apply(this, args);
-    };
+    }
+    Object.assign(wrapper, fn)
```
megawac: What is the purpose of this change?

aearly: So static properties attached to the `fn` get attached to the `wrapper`, namely the `PROMISE_SYMBOL` attached to `promiseCallback`s.

aearly:
@hargasinski Don't sweat it, after all, this project is run on donated time.

@megawac I hear your concerns about whether we should do this by default -- this is a rather significant change.  True, users can promisify methods, but I'm thinking about developer convenience.  Already, people are omitting the final callback on an Async method and being surprised a promise is not returned.  Several other popular libraries already work like this.

`async`/`await` is a very fundamental shift in how async code can work in JavaScript, and IMO it finally obsoletes callback-based approaches.  It feels weird to not give first-class support to it -- I feel we're somewhat obligated to because we control the package named `async` on npm.  Newer users of JavaScript are going to become more and more confused about the purpose of this library if we don't make some changes.

This next major version seems like a good point to throw more weight behind `async`/`await`-based approaches.

megawac:
How does this affect performance Alex? If its a significant hit my vote will be to seperate the asyncified and original methods into 2 files. I'd be completely fine with the async compatible version being our default export.

I know several people who use our library due to it's performance for intensive use cases so I'm adamant about maintaining good benchmarks :)

aearly:
There was no significant difference in the benchmarks on node 8 or 10.  I was concerned the extra wrapper would slow things down, but that is not the case.

hargasinski:
This is some really awesome work! The change look good to me.

Personally, I'm in favour of it. I think there has been enough interest/requests (#439, #956, [promise-async](https://github.com/Kikobeats/promise-async), [koa-async](https://github.com/eladnava/koa-async), [async-q](https://github.com/eladnava/koa-async)) to justify adding it. Also, looking over past issues, it seems that historically we've been against it because promises weren't implemented natively. Now that they are, and with significant parts of the community moving towards promise-based approaches, I think this change makes sense for us.

PR #1585, opened 2018-10-01
--save no longer needed

JoeCortopassi:
`--save` is on by default as of [npm 5](https://blog.npmjs.org/post/161081169345/v500), so `npm install aphrodite` is functionally equivalent to `npm install --save aphrodite` now


aearly:
:+1:

PR #1590, opened 2018-10-26
Set focus on search input on page load

davoam:
Currently when docs page is loaded focus is not set on search input. So a user has to click on the input to type. It is not convenient. 

Although the input has `autofocus` attribute it gets ignored, because of typeahead. 

I tried and tested solution from [here](https://github.com/twitter/typeahead.js/issues/444). It helped


aearly:
This is a great enhancement, but can you PR this to master?  `gh-pages` will get updated next time we publish.

davoam:
@aearly Hi, I did new PR to master branch https://github.com/caolan/async/pull/1591

Issue #1587, opened 2018-10-04
Feature request: reflect() should return values even when there is error


jazanne:
I have a function that will always return a "fallback" value to the callback, even if there's an error. I would really like reflect() / reflectAll() to capture this value.

For example:

```
function myFunc(callback){
    return callback('my error, 'my fallback value')
}

async.parallel(reflectAll({
   myval: myFunc
}), function(err, r){
    // r.myval.error = 'my error'
    // r.myval.value = 'my fallback value'
});
```

aearly:
Good point, we could easily return both the error and value in that case.

PR #1593, opened 2018-11-02
Reflects any values passed even if function has error

jazanne:
reflect/reflectAll return any "value" passed even if function also passes error. Includes tests.

For example:

```
function myFunc(callback){
    return callback('my error, 'my fallback value')
}

async.parallel(reflectAll({
   myval: myFunc
}), function(err, r){
    // r.myval.error = 'my error'
    // r.myval.value = 'my fallback value'
});
```

Addresses Issue #1587

### DIFF
--- a/lib/reflect.js
+++ b/lib/reflect.js
@@ -44,14 +44,18 @@ export default function reflect(fn) {
     var _fn = wrapAsync(fn);
     return initialParams(function reflectOn(args, reflectCallback) {
         args.push((error, ...cbArgs) => {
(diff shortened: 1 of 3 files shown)


aearly:
Thanks for this!  Can you test on node 8 or 10? -- there's a couple of tests that only run on those platforms that need updating as a result of this change.

aearly:
@jazanne There's still a little more work to be done here.

aearly:
Please re-open if you get this passing on node 8 and 10.

PR #1595, opened 2018-11-15
Use heap tree in priority queue

### LINKED ISSUE #1594: Use heap tree in priority queue
Hi people,

I've been using async for about some time. I needed use the priority queue recently but I disappointed by the fact that the implementation wasn't using a heap tree. I've seen [this](https://github.com/caolan/async/issues/1243#issuecomment-257472132) comment back in 2016. I don't know about your policy on contributions but I want to implement this. And I have some questions regarding it:

Heap tree is not stable by default. Meaning that the push order for the jobs having the same priority won't be preserved. To stabilize, another property might be added aside from **priority**, which would hold the push order for each job (call it **pushOrder**). And when doing comparisons in the heap, that **pushOrder** could be used for the jobs with the same priority. The real problem is that we need to hold a counter in the queue (or heap) which we increment each time the push method called. And that counter could overflow during a long term use of the queue. Can we ignore this? Or do you have a better idea?

Great library by the way.

yemreinci:
I implemented a stable heap tree. I replaced the DLL with the heap in priorityQueue. 

Let me know if you see anything wrong.

Fixes #1594

### DIFF
--- a/lib/internal/Heap.js
+++ b/lib/internal/Heap.js
@@ -1,0 +1,121 @@
+// Binary min-heap implementation used for priority queue.
+// Implementation is stable, i.e. push time is considered for equal priorities
+export default class Heap {
+    constructor() {
+        this.heap = [];
+        this.pushCount = Number.MIN_SAFE_INTEGER;
+    }
+
+    get length() {
+        return this.heap.length;
+    }
+
+    empty () {
+        this.heap = [];
+        return this;
+    }
+
+    percUp(index) {
+        let p;
+
+        while (index > 0 && smaller(this.heap[index], this.heap[p=parent(index)])) {
+            let t = this.heap[index];
+            this.heap[index] = this.heap[p];
+            this.heap[p] = t;
+
+            index = p;
+        }
+    }
+
+    percDown(index) {
+        let l;
+
+        while ((l=leftChi(index)) < this.heap.length) {
+            if (l+1 < this.heap.length && smaller(this.heap[l+1], this.heap[l])) {
+                l = l+1;
+            }
+
+            if (smaller(this.heap[index], this.heap[l])) {
+                break;
+            }
+
+            let t = this.heap[index];
+            this.heap[index] = this.heap[l];
+            this.heap[l] = t;
+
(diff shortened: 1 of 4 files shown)


yemreinci:
Here is the benchmark result for 50000 tasks:
priorityQueue(50000) v3.0.1-0 x 0.02 ops/sec ±0.00% (1 run sampled), 6.35e+4ms per run
priorityQueue(50000) current x 2.67 ops/sec ±9.51% (6 runs sampled), 375ms per run
current is faster

aearly:
Really cool work!  I'm a bit hesitant, because as you say, the prorityQueue sorting no longer becomes a stable sort.  It would be a subtle breaking change.  However, this feature could sneak in to the 3.0 release, since it is still in RC mode.

yemreinci:
@aearly, you might have misunderstood me. I said it wouldn't be stable if a plain heap was used. As you can see below, I create another property pushCount for objects pushed into the heap and use it to compare elements with the same priority. This pushCount property is unique for each element since the count is increased by 1 each time the push called. This costs another θ(n) space, but it is still same with the DoublyLinkedList. Because DLL creates next and prev properties whereas the Heap creates priority and pushCount. So I thought it would be alright to add another property.

pushCount is created here:
https://github.com/caolan/async/blob/e9676f14aac1de235524c2a15ffc8ee843bb6082/lib/internal/Heap.js#L50-L54
If the priorities are equal, element with the smaller pushCount is favored:
https://github.com/caolan/async/blob/e9676f14aac1de235524c2a15ffc8ee843bb6082/lib/internal/Heap.js#L113-L120

Anyways, it is stable now. There is just one little concern though. Heap.pushCount is beginning from MIN_SAFE_INTEGER and increases by one each time the push is called. That means after ~10^15 elements are pushed, it could break because of overflow. But I guess we can assume no user will push that much 😀.

https://github.com/caolan/async/blob/e9676f14aac1de235524c2a15ffc8ee843bb6082/lib/internal/Heap.js#L6


It wouldn't pass the tests if it wasn't stable, by the way.

PR #1597, opened 2018-11-20
Fixes #1582

### LINKED ISSUE #1582: `concat` and `groupBy` aren't cancelable
master

Node v10.5.0

**What did you do? Please include a minimal reproducable case illustrating issue.**
```js
async.groupBy([1, 2, 3], (item, callback) => {
    console.log(item);
    if (item === 2) {
        return callback(false);
    }
    callback(null, 'foo' + 1);
});
```

`groupBy` to stop running after the second iteration.

`groupBy` keeps running.

---
 In `concatLimit` and `groupByLimit` (and maybe others?), we check for an error, and if there is none, we pass `null` to the iteratee callback, which prevents the `false` value from propagating further. We just need to change
```js
  if (err) return iterCb(err);
  return iterCb(null, {key, val});
```
to
```js
  if (err) return iterCb(err);
  return iterCb(err, {key, val});
```
and add tests.

hargasinski:
This PR ensures that method pass the `err` onto `eachOfLimit`, so that they can be canceled properly. The one method I'm not quite sure about is `tryEach`. I'm assuming that since `retry` is cancelable, `tryEach` should be too?

Sorry if I went a little overboard with the tests, I ended up going down a rabbit hole :sweat_smile:. For the tests, I was trying to follow the coding style of the encompassing file, which is why some of the descriptions and styles vary. I figured it would be better to keep consistent logging and for future maintenance.

### DIFF
--- a/lib/concatLimit.js
+++ b/lib/concatLimit.js
@@ -26,7 +26,7 @@ function concatLimit(coll, limit, iteratee, callback) {
     return mapLimit(coll, limit, (val, iterCb) => {
         _iteratee(val, (err, ...args) => {
             if (err) return iterCb(err);
-            return iterCb(null, args);
+            return iterCb(err, args);
         });
     }, (err, mapResults) => {
         var result = [];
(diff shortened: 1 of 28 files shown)


hargasinski:
Thanks!

PR #1615, opened 2019-01-29
Fixes #1610

### LINKED ISSUE #1610: Documentation does not mention if async.eachSeries processes the array in order
I have a use case where I want to send a list of requests in order. I used async.eachSeries but one of my colleagues pointed out that the documentation does not mention if async.eachSeries processes the array in order or just one item at a time. 

After reading the code, my guess is it does. I maybe wrong, but the docs should mention this explicitly.

sam09:


hargasinski:
Thanks! The builds are failing, but it doesn't seem related to the PR...

@aearly you're more experienced with travis than me, have you seen these errors before?

sam09:
`sh -e /etc/init.d/xvfb start` fails. It is pre test command that is failing for some reason

aearly:
I believe I've fixed the issue with xvfb on master.

PR #1623, opened 2019-02-15
fix: priorityQueue drains while items still pending

andrebsguedes:
Priority queue should check if the underlying queue is idle before draining.

Currently the priorityQueue drains if an empty array of tasks is pushed even if other tasks are processing.

### DIFF
--- a/test/priorityQueue.js
+++ b/test/priorityQueue.js
@@ -239,5 +239,40 @@ describe('priorityQueue', () => {
             q.push('foo4', 1, () => {calls.push('foo4 cb');});
         });
     });
+    
+    it('should not call the drain callback if receives empty push and tasks are still pending', (done) => {
+        var call_order = [];
+
+        var q = async.priorityQueue((task, callback) => {
+            call_order.push('process ' + task);
+            callback('error', 'arg');
+        }, 1);
+        
+        q.push(1, 1, (err, arg) => {
+            expect(err).to.equal('error');
+            expect(arg).to.equal('arg');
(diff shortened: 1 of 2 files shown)


hargasinski:
Thanks for catching this! We have the same check in [queue](https://github.com/caolan/async/blob/dc9b3dea692343ec6688e98126cf5a2f661abc7e/lib/internal/queue.js#L29), so it makes sense to add it to the priorityQueue too.

Could you add a test?

aearly:
Can you write a test for this?

Also, are normal `queue`s susceptible to this?

andrebsguedes:
Glad to help! Added the test

test/priorityQueue.js
```diff
@@ -239,5 +239,40 @@ describe('priorityQueue', () => {
...
+        
+        expect(q.length()).to.equal(2);
+        
+        q.drain = function () {
+            expect(call_order).to.eql([
+                'process 1', 'callback 1',
+                'process 2', 'callback 2'
+            ]);
+            expect(q.concurrency).to.equal(1);
+            expect(q.length()).to.equal(0);
```
hargasinski:
```suggestion
            expect(q.length()).to.equal(0);
            expect(q.running()).to.equal(0);
```

Just one small change, `q.running()` should be checked too, as `q.length()` only gives the number of tasks waiting to be processed. Otherwise, it looks good, thanks!

andrebsguedes: Done!

hargasinski:
Thanks!

PR #1630, opened 2019-03-06
Add Azure Pipelines for CI (Windows, Linux, Mac)

pablonete:
Azure Pipelines gives open source projects 10 build pipelines with unlimited minutes for Linux, Mac, and Windows.

This PR adds support to run CI on all three platforms. We've been playing with the battery of tests and found them flaky on different OS. We've made some changes to tests trying to make them more stable, but based on the current approach (relying on setTimeout heavily), there's a big amount of randomness that makes hard to get a repeteable run in all OS.

We also made a fix in production code for an issue with serialization of arrow functions in Safari (see [eabb9f78](https://github.com/pablonete/async/pull/1/commits/eabb9f78ee00a9a81fc0d9012a2d0fca11c5d93c))

I set up my own pipeline [here](https://dev.azure.com/pablo-nunez/Runner/_build?definitionId=15). Here's a screenshot of the UI:
image

Azure Pipelines also has rich test reporting that's been set up and looks like this:
image

Let me know if there are any questions or suggestions. I'm ready to split this PR if that helps to integrate the changes on smaller bites. I'm a Software Engineer for Azure Pipelines and happy to help!


aearly:
Wow, this is an impressive amount of work.  Our tests really needed some love.

Is switching to Azure Pipelines necessary?  I wasn't planning on switching off Travis CI.  I am interested in testing in Edge and Safari, but was planning on using another service for that.  Sell me on Azure!

pablonete:
@aearly Thanks!

In this proposal, tests are run on Safari -so we caught a bug in arrow function serialization. We should be able to include Edge too, I'll give a try to karma-edge-launcher this week, it seems to be broken now: https://github.com/MicrosoftEdge/edge-launcher/issues/23

Azure Pipelines can be used for free in Open Source projects (10 parallel builds, unlimited CI/CD minutes), it supports the 3 main platforms in a common pipeline definition with aggregated results. It can be set up alongside with any other CI/CD tool.

aearly:
Hmnn, I wasn't able to set up an Azure Pipeline because I'm not technically an owner of this repo.  This is something @caolan would have to do.  However, I think it's way overdue for us to move this project into a github org so we can have multiple owners.  @Kikobeats set up https://github.com/async-js a few years back for us as part of an experiment.

caolan:
I've tried setting this up here for now: https://dev.azure.com/caolanmcmahon/async/_build?definitionId=1

pablonete:
It looks good, @caolan. It only seems missing the connection with coveralls. `COVERALLS_REPO_TOKEN` variable must be set in the Azure Pipeline so it's kept privately. Value is available in https://coveralls.io

jeremad:
Did you guys ever made that work?
https://stackoverflow.com/questions/61875369/karma-tests-on-edge-and-azure-pipelines

PR #1632, opened 2019-03-10
fix: Reflect passes value on error

### LINKED ISSUE #1587: Feature request: reflect() should return values even when there is error

aearly:
Fixes #1587

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,9 @@ node_js:
   - "8"
   - "10"
 
+services:
+  - xvfb
+
 matrix:
   include:
     - node_js: "10"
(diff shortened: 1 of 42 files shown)


aearly:
Nevermind, rebase really messed up the commits here.

PR #1641, opened 2019-05-19
BREAKING CHANGE: awaitable queues

### LINKED ISSUE #1586: Await-able queues
I've been thinking about how we might make queues work better with async/await.  It would be nice if certain operations returned promises, so they could be awaitable.  However, most of the callbacks used in queues are really more event-based, rather than a single task that resolves once.

For example, `q.push(task)` also accepts a callback.  We could make `push()` return a promise if you leave the callback off.  However, `push()` also accepts an array of items to push, in which case, the callback is called multiple times.  Promises can't resolve multiple times.

```js
const done = await q.push([foo, bar, baz]) // how should this resolve?
```

We could make it so it only resolves when all the tasks complete, but this complicates internal logic.

The handling of errors in this case would also be a bit strange.  The callback for `push()` is mainly useful for determining if an error occurred processing the item. If we return a Promise, suddenly that error becomes an unhandled rejection.  If we return a promise from `push()` suddenly a lot of existing code starts throwing unhandled rejections, as most people dont use the `push()` callback.  We could change it so the promise resolves with the error object in those cases.

There are several event callbacks you can register, `empty`, `saturated`, `error` etc.  It seems really difficult to make these awaitable in a useful way.  I've thought about changing the QueueObject API so that event  handlers are methods, rather than bare properties.

```js
q.error((err, task) => console.log(`Queue error: ${err.message} for ${task}`))

//rather than

q.error = (err, task) => console.log(`Queue error: ${err.message} for ${task}`)
```

In this case, we could make it so calling the method without passing a function returns a promise that resolves the next time the event triggers.

```js
const [err, task] = await q.error()
```

This is a bit clunky.  To repeatedly await errors, you'd have to use an infinite loop:

```js
var q = async.queue(taskFn)

const errorLoop = async () => {
  while(q.length) { // not a good condition to test for
    const [err, task] = await q.error()  // if no error occurs, this waits here forever
    console.log(err)
    q.push(task)
  }
}
errorLoop() // fire and forget, hopefully it exits when the queue drains.
```

It's also a bit complicated to implement internally.  We have to keep track of the  promises for each event, creating promise references for each and tracking them appropriately.

Returning a promise on the next firing of the event would be the most useful for the `drain` callback.  It makes the most common use of the queue work nicely in the context of an async function:

```js
async processItems(items) {
  const q = async.queue(processAsync, 5)
  q.push(items)
  await q.drain()
}
```

These changes add slightly  more utility to the queue in `async`/`await` but might be more trouble than they're worth.

-----

Another idea for the event-style callbacks is to return an Async Iterator:

```js
const iterator = q.error()

for await ([err, item] of iterator) {
  //...
}
```

These would still have to be wrapped in a fire-and-forget async function, unless you used one of the `for await` loops to drive the lifecycle of the queue within the main async function.  I think we would also have to assume that a `drain` event  means the queue has ended, thus also ending all the async iterators.

Lots of implementation complexity here as well.

aearly:
Closes #1586 

This changes the queue API a bit, allowing queue events to be awaited, as well as the results for an individual task.  You can now do:

```js
var q = async.q(async fn (task) { /* ... */ }, 10)

await q.push(task) // awaits until task has been processed
await Promise.all(q.push([1, 2, 3])) // awaits for all tasks to be processed

q.push(5) //set and forget, this won't reject if an error happens

await queue.drain() // probably the most useful thing to come out of this change
```

### DIFF
--- a/lib/internal/queue.js
+++ b/lib/internal/queue.js
@@ -147,12 +147,7 @@ export default function queue(worker, concurrency, payload) {
         },
         concurrency,
         payload,
-        saturated: eventMethod('saturated'),
-        unsaturated: eventMethod('unsaturated'),
         buffer: concurrency / 4,
-        empty: eventMethod('empty'),
-        drain: eventMethod('drain'),
-        error: eventMethod('error'),
         started: false,
         paused: false,
         push (data, callback) {
@@ -222,5 +217,28 @@ export default function queue(worker, concurrency, payload) {
             setImmediate(q.process);
         }
     };
+    // define these as fixed properties, so people get useful errors when updating
+    Object.defineProperties(q, {
+        saturated: {
+            writable: false,
+            value: eventMethod('saturated')
+        },
+        unsaturated: {
+            writable: false,
+            value: eventMethod('unsaturated')
+        },
+        empty: {
+            writable: false,
+            value: eventMethod('empty')
+        },
+        drain: {
+            writable: false,
+            value: eventMethod('drain')
+        },
+        error: {
+            writable: false,
+            value: eventMethod('error')
+        },
+    })
     return q;
 }
--- a/lib/internal/queue.js
+++ b/lib/internal/queue.js
@@ -3,8 +3,6 @@ import setImmediate from './setImmediate';
 import DLL from './DoublyLinkedList';
 import wrapAsync from './wrapAsync';
 
-const noop = () => {}
-
 export default function queue(worker, concurrency, payload) {
     if (concurrency == null) {
         concurrency = 1;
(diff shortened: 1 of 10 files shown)


lib/internal/queue.js
```diff
@@ -16,32 +14,68 @@ export default function queue(worker, concurrency, payload) {
...
+    const events = {
+        error: [],
+        drain: [],
+        saturated: [],
+        unsaturated: [],
+        empty: []
+    }
+
+    function on (event, handler) {
+        events[event].push(handler)
```
megawac: That's quite an interesting way of handling both promises and the current API. Is the promise based API useful for any of the events besides `drain`?

aearly: It's most useful for `drain`, but i can see utility coming from `unsaturated`, and making `error` reject.  I'm also happy making the interface uniform for all the event-style methods.

darksabrefr:
`await q.push(task)` is not safe for use for the moment because the promise returned by `q.push` never resolve nor reject if the queue worker throws an error. I understand that `await q.push(task)` cannot throw an error because of the backward compatibility of this method and its optional callback parameter (the push and forget feature). A `q.asyncPush(task)` method that reject its promise shouldn't be feasible ?

You can show the problem with this code (in node or browser without the require):
```javascript
let pushResult;

(async () => {
	const q = async.queue(async (task) => {
		throw new Error('Bad thing');
	});

	try {
		pushResult = q.push({some: 'data'});
		await pushResult;
		console.log('Error not catched');
	} catch (e) {
		console.log('Error catched');
	}
})();

setTimeout(() => {
	console.log('exit');
	console.log(pushResult);
}, 1000);
```

PR #1648, opened 2019-05-27
docs: use `docs/` folder, include links for every major version

### LINKED ISSUE #1583: Doc versioning and other `v3` doc issues
Right now, when the docs are generated, the new version replaces the previous one. When `v3` comes out, we will need both the `v3` and `v2` docs on the site, and a way to update them individually. Depending on the amount of history we want, this should be fairly straight forward with a script that detects the version being generated from the `package.json`. Should we just have the latest docs for  each major version (`v2`, `v3`), each minor version (`v2.5`, `v2.6`, `v3.0`, `v3.1`, etc...) or each patch version (`v2.6.0`, `v2.6.1`, `v3.0.0`, `v3.0.1`, etc...)? I'm leaning towards having the docs for all the patch versions available on the site and being able to switch between them with the dropdown at the top.

Are there any other doc issues that need to be addressed for `v3`?

aearly:
Closes #1583 

Now we can publish the docs at will, as well has have a frozen version of the 2x docs.


aearly:
Ah, this will require @caolan to enable that setting for the project.  I can't change it from the `gh-pages` branch strategy.

PR #1652, opened 2019-05-29
Azure pipelines badges

pablonete:
Added badges for Pipelines and for each platform in a table.

**Note:** Windows and Linux builds are red due to the lack of COVERALLS_REPO_TOKEN variable.

**Note 2:** libhive.com SSL cert is wrong/expired. I'm ignoring it.

See badges preview: https://github.com/pablonete/async/blob/azure-pipelines-badges/README.md


aearly:
@caolan we need that env var fixed :point_up:

PR #1682, opened 2019-09-05
Update whilst.js.html

eesc88:
A comma cannot be placed in parenthesis.

PR #1686, opened 2019-09-15
Update whilst.js.html

alokadhao20:
This Semicolon(;) should not be present for error-free execution of the code.


aearly:
See #1682

PR #1689, opened 2019-09-25
Fix docs url

karataev:


aearly:
There's a redirect already in place for this.  When v4 is released, I'll simply update it to point to `/v4/`.  `https://caolan.github.io/async/` _is_ the canonical URL for docs.

karataev:
Well, for me 404 page is opened instead of redirect 🤷‍♂️

aearly:
Ah, I forgot that this was an HTML-based redirect rather than a rewrite-style redirect.

PR #1690, opened 2019-10-04
Check for name or message of Error before wrap in new Error

svedbg:
Relying on `err.message` is not great, because in some cases it might be undefined. That makes handling of errors really hard.

### DIFF
--- a/docs/v2/asyncify.js.html
+++ b/docs/v2/asyncify.js.html
@@ -63,15 +63,15 @@
 
 <div id="main">
     <div id="main-container" data-spy="scroll" data-target="#toc" data-offset="50">
-        
+
         <h1 class="page-title">asyncify.js</h1>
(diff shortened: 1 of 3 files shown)


aearly:
I'm :-1: on this.  An error object without a `message` is really non-standard.

svedbg:
Please exmplain why you think `error` object without message is non-standard? In MDN https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error `message` is optional, that means we really can have as standard error objects without message set.  
Also if you take a look at the test I gave, the error is constructed using:
```
        class CustomError extends Error {
            get name() {
                return 'CustomError';
            }
        }
```
That's simple and elegant way for defining errors which can be thrown from the async blocks and handled later in the code.

Issue #1700, opened 2020-01-13
Don't redefine property "name"


binarykitchen:
Happens in your latest version:

```
Uncaught TypeError: Cannot redefine property: name
```

Debugged it in the raw, compiled ES5 code and found the cause:

```
return Object.defineProperty(awaitable,"name",{value:"awaitable(".concat(asyncFn.name,")")})
```

The name property has been set without `configurable: true` and is being redefined again when another code requires the same file. I'd consider adding `configurable: true`. See: https://stackoverflow.com/questions/25517989/why-cant-i-redefine-a-property-in-a-javascript-object

(Sorry for not following the issue template but this is simple.)

binarykitchen:
Can somebody look at this please? It's getting serious, happening every hour on my production site.

Or would you accept a PR of mine?

binarykitchen:
Ping @karataev @aearly ???

karataev:
@binarykitchen Sorry, I'm not a maintainer and can't help you

binarykitchen:
@karataev but i see you made some recent commits?

karataev:
@binarykitchen As far as I remember, it was just a pull request with documentation fix

binarykitchen:
okay, i hope @aearly will comment

binarykitchen:
Alright, here's a PR https://github.com/caolan/async/pull/1701

aearly:
Published in `v3.1.1`

binarykitchen:
@aearly Unfortunately it is still happening in the latest version. Can you reopen this ticket?

binarykitchen:
I guess this happens with some exotic user agents like

```
"userAgent": "Mozilla/5.0 (Linux; Android 4.2.1; en-us; Nexus 5 Build/JOP40D) AppleWebKit/535.19 (KHTML, like Gecko; googleweblight) Chrome/38.0.1025.166 Mobile Safari/535.19"
```

binarykitchen:
Also I do not recommend having code to rely on function names because of minifiers and obfuscators trying to reduce size of reoccuring strings.

jpcodr:
Hi, i have the same issue.  Some workaround? @binarykitchen
In my case occurs on a cordova app with Android 5.0:
```json
[...]

PR #1701, opened 2020-01-22
Allow redefining name property in awaitable

binarykitchen:
Solves https://github.com/caolan/async/issues/1700


aearly:
Will publish this later today

PR #1702, opened 2020-01-24
test to prove bug with await expectation on queue.drain

JefStat:
I was expecting the drain function to always return the same promise. But since a promise cannot resolve more then once it would break the expectation of how you use the queue. Especially in the case where the queue would start and stop. For instance you add more items to the queue after it has drained once. 

@aearly Pinging you since you were the original implementor of the awaitable feature

### DIFF
--- a/lib/internal/queue.js
+++ b/lib/internal/queue.js
@@ -132,16 +132,17 @@ export default function queue(worker, concurrency, payload) {
 
     const eventMethod = (name) => (handler) => {
         if (!handler) {
-            return new Promise((resolve, reject) => {
+            const p = new Promise((resolve, reject) => {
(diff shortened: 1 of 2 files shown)


JefStat:
The  test that fails is commented at flaky         `this.retries(3); // test can be flakey` The test passes locally

aearly:
I'm not really sure what the error/bug you're reporting is.  The change you made also doesn't really make sense -- why would we call `_maybeDrain` after creating promises for unrelated events?

JefStat:
I have edge case based a race condition where I call queue.drain(); which returns a promise which will never be resolved since there will be no more items added to the queue to invoke the drain.

PR #1705, opened 2020-02-18
fix: handle ES2017 Async Functions in `async.apply()`

aryehb:
Previously, `async.apply()` always returns a regular function, even if an async function was passed in. This meant, for example, that `async.apply()` would not be able to be used in other methods without
wrapping in `async.asyncify()`.

This fix checks what type of function was passed in, and returns the same type.

### DIFF
--- a/lib/apply.js
+++ b/lib/apply.js
@@ -1,4 +1,5 @@
 import { isAsync } from './internal/wrapAsync';
+import asyncify from './asyncify';
 
 /**
  * Creates a continuation function with some arguments already applied.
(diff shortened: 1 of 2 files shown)


aearly:
I'm hesitant to do this, since we currently don't use the `async` keyword anywhere in the library.  Including this will cause projects that use Async to now do more transpilation and pull in things like `regenerator` in a lot of downstream projects.

aryehb:
I understand your concern. I changed the fix to wrap with `asyncify` instead of an async function.

aearly:
Hmnn, I think it would be surprising if `apply` turned an `async` function to a callback-style function.  Maybe this should wait until the next major version.

aryehb:
>Hmnn, I think it would be surprising if apply turned an async function to a callback-style function.

That's a good point. I was only thinking in terms of using it along with `async.waterfall()`.

>Maybe this should wait until the next major version.

I'll close this for now, and it can be revisited if `async` is added at a later date.

aearly:
FWIW, the built in `asyncFn.bind(null, arg1, arg2...)` preserves the `async`-ness of the returned function, as well as things like `async arg3 => asyncFn(arg1, arg2, arg3)`.

aryehb:
That's what I switched to using in the meantime, but I thought it would be nice if `async.apply()` would support it out of the box.

Issue #1609, opened 2019-01-13
Documentation needs better examples


joshua-mbg:
Consider this example for map
`async.map(['file1','file2','file3'], fs.stat, function(err, results) {
    // results is now an array of stats for each file
});`

This basically tells me nothing except syntax. In most cases people aren't going to pass `fs.stat`. The examples should include use of all possible options. All examples for all functions and methods need to be updated. They are too simplistic

aearly:
I agree, we need much better docs.  I'd also like to change many of the examples (or add mroe examples) to use async/await.  However, this is something that takes time I don't have at the moment.

mayur-gupta:
Maybe I can help.

poloyc:
How can I help with examples?

aearly:
- Examples with more `async`/`await`, less with callbacks
- More concrete examples from real-world applications, less `foo`/`bar`
- Things that show the "sweet spot" for Async -- things that are hard to do without the library
- Examples the are accessible to newcomers to Node/JS that show how parameters flow through
- More comprehensive examples showing more features of individual methods

There are `@example` blocks in the JSDoc comments for most methods.  Go nuts!

caub:
note for that particular example, you could do
```js
await async.map(['file1','file2','file3'], fs.promises.stat);
// or without async lib
await Promise.all(['file1','file2','file3'].map(f => fs.promises.stat(f)))
```
or if you want an object as result
```js
await async.auto({
  file1: async()=>fs.promises.stat('file1'),
  file2: async()=>fs.promises.stat('file2')
  file3: async()=>fs.promises.stat('file3')
})
// or without async lib
Object.fromEntries(
  await Promise.all(['file1','file2','file3'].map(async name => [name, await fs.promises.stat(name)]))
)
```

flatcoding:
I'm in particular looking for a (simple) Promise-based example of async.mapLimit, even without async/await. Haven't found any non-callback example so far at all.

caub:
@flatcoding https://github.com/caub/misc/blob/master/utils/promise-concurrent.js, if you don't want `await`, you can replace it by a recursive function

flatcoding:
Hi Cyril @caub,
thanks for your reply. I'm not particularly against 'await', rather for the sake of 'good examples' for the 'async' module I'd like to start with and see an implementation with basic Promise handling, including .then .catch.
I honestly struggle using async with Promises in general, so before using async/await I'd just like to start with step 1) ...
[...]

PR #1726, opened 2020-10-05
Enhance examples for Collections methods.

romanbalayan:
Partial documentation update for #1609 

Updated collections method examples, including usage for callbacks, promises and async/await


aearly:
Hi, thanks for this!

I agree with @caub in #1609 -- the promise examples aren't that useful -- I think we should just have callback style, and `async`/`await` style, and also examples that use `async` iteration functions.

gkatsanos:
Given the work @romanbalayan did for this, isnt' there a possibility to merge the PR even partially? It's a pity when contributors come with work and are said "well this is excellent but pretty much rewrite it". ?

PR #1730, opened 2020-10-25
Fix typo

tmacro:
`unshirtAsync` => `unshiftAsync`


aearly (changes requested):
Can you update the source for these docs in lib/queue.js?

tmacro:
sure thing

tmacro:
@aearly should I remove the fix in `docs/v3` as upon closer inspection it looks like they're generate from the source?

aearly:
You can run  `make doc` and it will regenerate.

aearly:
Re-open this when you fix `lib/queue.js`

PR #1752, opened 2021-06-19
change argument position

chdnitinsharma:


chdnitinsharma:
Hi @aearly , If we do not swap the argument then it show `callback is not a function`.

PR #1761, opened 2021-08-02
use queueMicrotask instead of setTimeout in browsers to avoid browser throttling

jensengar:
Since `setTimeout` get throttled by browsers and `setImmediate` doesn't actually exist in any browsers, we should use `queueMicrotask` which works very similarly to the intention of `setImmediate`.

I have an app that uses stanza.io which relies on async's `priorityQueue`. It was found that `priorityQueue` is getting throttled because of its use of `setTimeout(fn, 0)`. Most browsers support `queueMicrotask`, but for those who don't, we will fallback.

### DIFF
--- a/lib/internal/setImmediate.js
+++ b/lib/internal/setImmediate.js
@@ -1,6 +1,7 @@
 'use strict';
 /* istanbul ignore file */
 
+export var hasQueueMicrotask = typeof queueMicrotask === 'function' && queueMicrotask;
 export var hasSetImmediate = typeof setImmediate === 'function' && setImmediate;
 export var hasNextTick = typeof process === 'object' && typeof process.nextTick === 'function';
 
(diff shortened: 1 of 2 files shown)


djhouseknecht:
For reference: 
* [Here is the article explaining Chrome's timer throttling](https://developer.chrome.com/blog/timer-throttling-in-chrome-88/)
* Here is the stanza repo: https://github.com/legastero/stanza

sauravazad (approved):

jensengar:
@sauravazad thanks for the approval. Is there anyone else who needs to approve this? And who has merge/release rights for this?

aearly:
I'll merge this later today.

aearly:
Published in v3.2.1

jensengar:
Thanks!

PR #1767, opened 2021-11-05
Fix an inefficient regex in autoInject

meekdenzo:
This fixes the issues highlighted in https://lgtm.com/projects/g/caolan/async/snapshot/f7b5d714b21bada7bd7a39b75e238bedcd162853/files/lib/autoInject.js?sort=name&dir=ASC&mode=heatmap#L12


Trott:
This is open source and no one owes anybody anything, but is there anything that might encourage a review of this PR?

aearly (changes requested):
Can you add a test to make sure comments in argument definitions are still properly stripped?  Both `(inline /* remove me */) => {}` and 
```
(
  // remove this
  singleLine
) => {}
```

lib/autoInject.js
```diff
@@ -6,7 +6,7 @@ var FN_ARGS = /^(?:async\s+)?(?:function)?\s*\w*\s*\(\s*([^)]+)\s*\)(?:\s*{)/;
 var ARROW_FN_ARGS = /^(?:async\s+)?\(?\s*([^)=]+)\s*\)?(?:\s*=>)/;
 var FN_ARG_SPLIT = /,/;
 var FN_ARG = /(=.+)?(\s*)$/;
-var STRIP_COMMENTS = /((\/\/.*$)|(\/\*[\s\S]*?\*\/))/mg;
+var STRIP_COMMENTS = /(\/\*(?:[^/]|\/(?!\*))*\*\/)|([^:]|^)\/\/.*$/mg;
```
Trott, line 9: Having looked more closely at the code, I think trying to allow for URLs with the `:` should be removed. This regex is only ever used on the string representation of code representing function parameters. The URL case will never come up. (And if it did, relying on a colon as the signifier for a URL would be unreliable.)

```suggestion
var STRIP_COMMENTS = /(\/\*(?:[^/]|\/(?!\*))*\*\/)|\/\/.*$/mg;
```

test/autoInject.js
```diff
@@ -4,6 +4,36 @@ var {expect} = require('chai');
...
+                return a;
+            }
+        })).to.deep.eql(
+            `(inline ) => {return a}` +
+            `(
+                
+                singleline
+            ) => {return a}`
+        )
+    });
```
Trott, line 35: Here is a test that will fail if either halves of the regular expression are not working:

```suggestion
    it('should properly strip comments in argument definitions', (done) => {
        async.autoInject({
            task1: function(task2, /* ) */ callback) {
                callback(null, true);
            },
            task2: function task2(task3 // )
                ,callback) {
                    callback(null, true);
            },
            task3: function task3(callback) {
                callback(null, true);
            }
        },
        (err, result) => {
            expect(err).to.eql(null);
            expect(result).to.deep.eql({task1: true, task2: true, task3: true});
            done();
        });
    });
```

meekdenzo: Thank you for your input @Trott 
This looks so much better!

meekdenzo:
@aearly how are we looking here?

test/autoInject.js
```diff
@@ -4,6 +4,35 @@ var {expect} = require('chai');
...
+                callback(null, true);
+            }
+        },
+        (err, result) => {
+            expect(err).to.eql(null);
+            expect(result).to.deep.eql({task1: true, task2: true, task3: true});
+            done();
+        });
+    });
+
```
Trott, line 35: It might be better to move the new tests near the end of the file, or at least somewhere after "basics". I would expect the basic tests to be run before anything checking the efficiency of a regular expression or edge cases involving comments.

meekdenzo: That's a good point @Trott I'll make the changes

test/autoInject.js
```diff
@@ -224,4 +224,33 @@ describe('autoInject', () => {
             done()
         })
     })
+
+        it('should not be subject to ReDoS', () => {
```
Trott, line 228: Is this line intended too far? (I'm surprised if there isn't a lint rule for this.)

aearly (approved):

PR #1769, opened 2021-11-10
fix: Explicit filepath

jimmywarting:
You are not truly using ESM unless you are not specifying the extension.
you are compiling to commonjs after all...

remote resolver don't have access to the filesystem and can't resolve `./foo` to `./foo.js` or `./foo/index.js`

https://nodejs.org/docs/latest-v17.x/api/esm.html#mandatory-file-extensions

the only thing i think you have left to do is adding `"type": "module"` in package.json and stop generating cjs/umd and just output esm, but this is up to you - at least now this is more cdn friendly


aearly:
Thanks.  ESM support has changed out from underneath us since we did the initial switch to ESM.

PR #1772, opened 2021-11-16
Housekeeping

jimmywarting:
When installing all dependencies then it yields some deprecation warnings. some packages wasn't even being used.
Where going to fix some of it but ended up doing something else instead.

---

Where thinking of removing `fs-extra` also but though if i should ask first...
it seems to only be used in development to fix some docs and reading some files
i where thinking that it wouldn't be so bad if it used nodes native `fs` after all 
And b/c it's dev only and mostly one time thing. i think it wouldn't be nice to use the sync api instead to remove some "callback-hell" stuff and avoid the need of async and or promises.
the sync methods are faster after all but they are blocking...

### DIFF
--- a/lib/seq.js
+++ b/lib/seq.js
@@ -26,7 +26,7 @@ import { promiseCallback, PROMISE_SYMBOL } from './internal/promiseCallback.js'
  * app.get('/cats', function(request, response) {
  *     var User = request.models.User;
  *     async.seq(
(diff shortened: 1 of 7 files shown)


aearly (approved):
Thanks for this!  The docs generation scripts could use a bit of love as well.  They were written when `async` / `await` support wasn't that great in Node.

jimmywarting:
would it be alright to use sync version instead?

PR #1783, opened 2022-01-10
chore: add Dependabot for version updates

nschonni:
Now the CI is testing across platforms, Dependabot can be used to start testing dependdency updates

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,12 @@
+version: 2
+updates:
+  - package-ecosystem: npm
+    directory: '/'
+    schedule:
+      interval: weekly
+    open-pull-requests-limit: 99
+  - package-ecosystem: github-actions
+    directory: '/'
+    schedule:
+      interval: weekly
+    open-pull-requests-limit: 99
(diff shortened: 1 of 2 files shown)


aearly:
Personally, I find dependabot annoying.  I'd prefer to do this kind of thing manually a few times a year at most, especially since we only have dev dependencies.

.github/dependabot.yml
```diff
@@ -0,0 +1,12 @@
+version: 2
+updates:
+  - package-ecosystem: npm
+    directory: '/'
+    schedule:
+      interval: weekly
```
nschonni, line 6: Could move this to monthly to reduce the noise

aearly: Maybe bi-annually?  All deps are devDeps, so it's not critical to keep everything up to date.

aearly (changes requested):

.github/dependabot.yml
nschonni:
```suggestion
      interval: monthly
```

nschonni:
I bumped it up to Monthly, but that's the make interval right now

PR #1786, opened 2022-01-10
chore: upgrade to ESLint 8

nschonni:
- New babel parser package
- Drops (dev) support for Node 10

Expect the Azure DevOps Node 10 build to fail


nschonni:
Not sure if it makes sense to drop the Azure DevOps builds now or not

PR #1789, opened 2022-01-23
docs: add missing return definitions for awaitable functions [fixes #1777]

### LINKED ISSUE #1777: Feature Request: waterfall to return promise when callback not defined
[Currently](https://caolan.github.io/async/v3/docs.html#waterfall), the `waterfall` method expects a callback for the result.

It would be great if `waterfall` method returns a promise when the callback is not provided, like other methods (eg: [`parallel`](https://caolan.github.io/async/v3/docs.html#parallel), [`series`](https://caolan.github.io/async/v3/docs.html#series))

Workaround:

```ts
import { promisify } from "util";
import { waterfall, constant } from "async";

const waterfallP = promisify(waterfall);

const add3 = async (arg1: number): Promise<number> => {
  return Promise.resolve(arg1 + 3);
};

const mul4 = async (arg1: number): Promise<number> => {
  return Promise.resolve(arg1 * 4);
};

const result = await waterfallP([
  constant(1),
  add3,
  mul4
]);

// result = 16
```

hargasinski:
Hey! It's been a while since I contributed to the repo, so I thought I would start with a couple of quick doc updates.

This PR adds the missing return definitions for `race` and `waterfall` and updates the `detect*` docs to match the style of all the other functions.

### DIFF
--- a/lib/detect.js
+++ b/lib/detect.js
@@ -27,7 +27,7 @@ import awaitify from './internal/awaitify.js'
  * Result will be the first item in the array that passes the truth test
  * (iteratee) or the value `undefined` if none passed. Invoked with
  * (err, result).
- * @returns A Promise, if no callback is passed
+ * @returns {Promise} a promise, if a callback is omitted
  * @example
  *
  * // dir1 is a directory that contains file1.txt, file2.txt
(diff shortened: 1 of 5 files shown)


aearly (approved):
Good to see you again, @hargasinski !

PR #1790, opened 2022-01-24
update priorityQueue functionality to match queue [fixes #1725]

### LINKED ISSUE #1725: PriorityQueue push doesn't wait for result
What version of async are you using?
^3.2.0
Which environment did the issue occur in (Node version/browser version)
Node: >=12
npm: 6.13.4

What did you do? Please include a minimal reproducable case illustrating issue.
Maybe I'm missing something but I thought that priorityQueue and queue have same behaviour
but if I run the following code, the output is 1
```js
 const queue = async.queue( async (task) =>{
   return Promise.resolve(1)
 })          
const res = await queue.push({})
console.log(res)
```

instead if I run the following code the output is undefined
```js
const queue = async.priorityQueue( async (task) =>{
  return Promise.resolve(1)
})
const res = await queue.push({},1)
console.log(res)
```

hargasinski:
This PR updates the `priorityQueue` behaviour to match the `queue` implementation:
* better supports promises in `priorityQueue` (see #1725).
* supports `pushAsync` and removes `unshiftAsync`.
* fixes a bug where `drain` wasn't being called when an empty task was pushed.
* eliminates duplicate code by keeping all of the `queue` logic in `queue`. The one issue with this is it now iterates over the `tasks` twice, once in `priorityQueue` and once in `queue`. Let me know if there's a better way you can think of.

I also added a few tests from `queue` to `priorityQueue` to make sure the functionality matches.

### DIFF
--- a/test/priorityQueue.js
+++ b/test/priorityQueue.js
@@ -16,13 +16,13 @@ describe('priorityQueue', () => {
         q.push(1, 1.4, (err, arg) => {
             expect(err).to.equal('error');
             expect(arg).to.equal('arg');
-            expect(q.length()).to.equal(2);
+            expect(q.length()).to.equal(3);
             call_order.push('callback ' + 1);
         });
         q.push(2, 0.2, (err, arg) => {
             expect(err).to.equal('error');
             expect(arg).to.equal('arg');
-            expect(q.length()).to.equal(3);
+            expect(q.length()).to.equal(4);
             call_order.push('callback ' + 2);
         });
         q.push(3, 3.8, (err, arg) => {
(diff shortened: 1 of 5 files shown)


hargasinski:
The failed tests don't seem to be related to this PR, tests for other functions are timing out.

test/priorityQueue.js
```diff
@@ -31,26 +31,25 @@ describe('priorityQueue', () => {
             expect(q.length()).to.equal(0);
             call_order.push('callback ' + 3);
         });
-        q.push(4, 2.9, (err, arg) => {
```
aearly, line 34: The expected lengths change, due to pushing this array, correct?

aearly (commented):
Thanks for this, this has been a missing feature for a while.  Just a small question.

test/priorityQueue.js
hargasinski: Yes sir, we don't test pushing an array in any of the other `priorityQueue` tests, so I thought it would be useful to add.

hargasinski:
@aearly is this good to go?

aearly (approved):

hargasinski:
I'm running into an issue with the build, looks like somewhere along the way one of `babel-minify`'s dependencies got updated and it's now throwing an error during minification. I'll look more into it this weekend and publish `v3.2.4` with the fix from this PR.

hargasinski:
@aearly published `v3.2.4` with this fix! Sorry for the delay, things have been busy on my end. I was hoping `babel-minify v0.5.2` would fix the minification issue, but it didn't. I was able to work around it to get the minified file built and the patch published though.

PR #1818, opened 2022-02-14
don't schedule another drain call if one is running [fixes #1791]

### LINKED ISSUE #1791: Pushing an empty task onto the queue in q.drain results in an infinite loop.
**What version of async are you using?** `v3.2.3`

**Which environment did the issue occur in (Node/browser/Babel/Typescript version)** node & browser

**What did you do? Please include a minimal reproducible case illustrating issue.**
```js
const q = async.queue(() => {})
q.drain(() => { q.push([]) })
q.push([])
```

The `queue` doesn't schedule another `drain` call if one is running.

The `queue` goes into an infinite loop. 

I'm not sure this is necessarily a bug as `queue` is working as specified - if an empty task is pushed and it's `idle`, it drains. I only noticed it because in the `priorityQueue` tests, we have a test for this case, see [tests/priorityQueue#L52](https://github.com/caolan/async/blob/master/test/priorityQueue.js#L52). With the changes in #1790, the `priorityQueue`behaves the same way as `queue` and goes into an infinite loop. We could add a check that if a `drain` call is running, not to schedule another one similar to what we do with `q._process`. Thoughts?

hargasinski:
This PR adds checks to `queue` to ensure that if `drain` is running/scheduled to run, another `drain` isn't called. It prevents an infinite loop in the case of: 
```js
q.drain(() => q.push([]))
```

### DIFF
--- a/lib/internal/queue.js
+++ b/lib/internal/queue.js
@@ -115,8 +115,10 @@ export default function queue(worker, concurrency, payload) {
                 trigger('unsaturated')
             }
 
-            if (q.idle()) {
+            if (q.idle() && !drainScheduled) {
+                drainScheduled = true
                 trigger('drain')
+                drainScheduled = false
             }
             q.process();
         };
--- a/lib/internal/queue.js
+++ b/lib/internal/queue.js
@@ -45,6 +45,7 @@ export default function queue(worker, concurrency, payload) {
     }
 
     var processingScheduled = false;
+    var drainScheduled = false;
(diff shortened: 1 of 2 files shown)


hargasinski:
@aearly after thinking about this more, I don't think it's something we need to worry about unless you have a cleaner fix. This feels janky. It's been in `queue` without issues, so I doubt it's causing any issues/unexpected behaviours.

Issue #1729, opened 2020-10-18
queue.drain() invoked even when there are pending tasks


Heedster:
**What version of async are you using?**
3.2.0

**Which environment did the issue occur in (Node/browser/Babel/Typescript version)**
Babel. I used the async package, but the bug seems to exist in async-es as well. 

**What did you do? Please include a minimal reproducible case illustrating issue.**
push an empty array, and then array(s) with task, and then wait for queue to drain.

Here is a sample code:
```js
import { queue } from "async-es";

const createTask = () => {
  return {
    doIt: () => {
      return new Promise((r) => setTimeout(r, 2000));
    }
  };
};

const createQueueAndProcess = async () => {
  const q = queue(async (task, callback) => {
    await task.doIt();
    console.log("task done");
    callback();
  });

  //Comment the empty array push line below
  // and see the issue go away
  q.push([]);
  q.push([createTask()]);

  await q.drain();
  console.log("all tasks completed");
};

createQueueAndProcess();
```

Code sandbox link: https://codesandbox.io/s/async-queue-issue-mqxo5

**What did you expect to happen?**
Complete the one task that was pushed and then call drain. So the console will say

```
task done
all tasks completed
```

**What was the actual result?**
drain is exited (event emitted) before the task getting completed. 

```
all tasks completed
task done
```

If you remove that empty array, things get fixed. In real world, I got this issue when I was calling functions for getting tasks to upload any pending files to the server, and some of the functions returned empty arrays. 

I did some digging around the internal/queue.js, and I think the issue is because of _maybeDrain and the setImmediate inside it. I understand the reason behind the code, but this is what is likely happening
- pushed empty array
- we are not doing anything, we should emit drain, schedule it.
- actual things pushed. (setImmediate still have not emitted the drain)
- this time we are not emitting coz we have actual work. 
- waiting for drain. 
- setImmediate now sees next tick, and calls the drain, which was supposed to be emitted long before.

der-kommissar:
I think I ran into something similar. In my case, it works when I run from source, but when I pack the module with `npm pack` and run the executable, it does not work.

Here's my test code:
```TypeScript
import * as Async from 'async';

class DriverQ {
    public async queueTasks (tasks: string[]): Promise<string[]> {
        const results: string[] = [];
        const queue = Async.queue(async (task: string, callback) => {
            console.log(`Queue driver for task ${task}`);
            const res = await this.runTask(task);
            results.push(res);
            callback();
        }, 2);
[...]

PR #1871, opened 2022-10-09
Update queue.js

heartforit:
This fixes an bug: Drain function got called even when some tasks still ongoing at the queue. Fix proposal is due to https://github.com/caolan/async/issues/1729 (@Domiii) Tested by me locally. Works great.


aearly:
You will need to fix the tests.

heartforit:
I won't best. I cancel my pull request. Tought will be esay and fast but it will not.

heartforit:
I will not fix this. I will switch to another option.

Domiii:
@aearly @heartforit Fixed my snippet. See original issue: #1729

PR #1924, opened 2023-11-06
Update changelog for v3.2.5

Mathias-S:
Seems to have been forgotten for the 3.2.5 release. Keeping the changelog updated is very useful to quickly review what has changed so we don't need to dig around in commit logs.


aearly:
Thanks

Issue #1975, opened 2024-07-02
Async 3.2.5 is vulnerable to ReDoS (Regular Expression Denial of Service) CVE-2024-39249


dvasilen:
Please assess and address the CVE-2024-39249 in Async 3.2.5

## CVE-2024-39249 - Medium Severity Vulnerability
<details><summary><img src='https://whitesource-resources.whitesourcesoftware.com/vulnerability_details.png' width=19 height=20> Vulnerable Library - <b>async-3.2.5.tgz</b></p></summary>

<p></p>
<p>Library home page: <a href="https://registry.npmjs.org/async/-/async-3.2.5.tgz">https://registry.npmjs.org/async/-/async-3.2.5.tgz</a></p>
<p>Path to dependency file: /package.json</p>
<p>Path to vulnerable library: /package.json</p>
<p>

Dependency Hierarchy:
  - :x: **async-3.2.5.tgz** (Vulnerable Library)
<p>Found in base branch: <b>master</b></p>
</p>
</details>
<p></p>
<details><summary><img src='https://whitesource-resources.whitesourcesoftware.com/medium_vul.png?' width=19 height=20> Vulnerability Details</summary>
<p>  
  
Async <= 2.6.4 and <= 3.2.5 are vulnerable to ReDoS (Regular Expression Denial of Service) while parsing function in autoinject function.

<p>Publish Date: 2024-07-01
<p>URL: <a href=https://www.mend.io/vulnerability-database/CVE-2024-39249>CVE-2024-39249</a></p>
</p>
</details>
<p></p>
<details><summary><img src='https://whitesource-resources.whitesourcesoftware.com/cvss3.png' width=19 height=20> CVSS 3 Score Details (<b>5.5</b>)</summary>
<p>

Base Score Metrics:
- Exploitability Metrics:
  - Attack Vector: Local
  - Attack Complexity: Low
  - Privileges Required: None
  - User Interaction: Required
  - Scope: Unchanged
- Impact Metrics:
  - Confidentiality Impact: None
  - Integrity Impact: None
  - Availability Impact: High
</p>
For more information on CVSS3 Scores, click <a href="https://www.first.org/cvss/calculator/3.0">here</a>.
</p>
</details>
<p></p>

aearly:
This seems incredibly unlikely to be exploitable, except in development. If a user of Async was eval()ing user input and passing functions to `autoInject`, they would have bigger problems.

PPKath-1611:
[...]

PR #1980, opened 2024-07-19
Fix ReDoS

kevinbackhouse:
This fixes the ReDoS issue that was reported in #1975.

I couldn't find a way to fix the ReDoS without making minor changes to the behavior of the regex, so I've split this PR into two commits to show what I've done. The first commit simplifies the regex, but in a way that will make it match a superset of the strings that it matched before. The second commit fixes the ReDoS (in both regexes) without changing their behavior.


aearly:
Thanks for fixing this.  The slightly expanded regex is totally fine.

zteric:
Hi @aearly , can we expect a quick release for this fix?

aearly:
Published in 3.2.6.

